PR #10, opened 2019-12-26
EPMDJ-2105. Fixed a change Header mapping setting

ViktoriiaLebedeva:


core/src/main/kotlin/com/epam/drill/endpoints/AgentManager.kt
```diff
@@ -6,6 +6,7 @@ import com.epam.drill.agentmanager.*
 import com.epam.drill.api.*
 import com.epam.drill.common.*
 import com.epam.drill.endpoints.agent.*
+import com.epam.drill.endpoints.openapi.*
```
viktorbll: Why this import?

ViktoriiaLebedeva: This is necessary to use SystemSettings, which are a function parameter.
And due to the fact that now we write all the imports through *, I added the import exactly like this

core/src/main/kotlin/com/epam/drill/endpoints/openapi/DrillAdminEndpoints.kt
```diff
@@ -295,10 +295,4 @@ class DrillAdminEndpoints(override val kodein: Kodein) : KodeinAware {
...
     }
-}
-
-@Serializable
-data class SystemSettings(
-    val packagesPrefixes: List<String> = emptyList(),
-    val sessionIdHeaderName: String = ""
-)
\ No newline at end of file
+}
```
viktorbll: newline!

core/src/main/kotlin/com/epam/drill/dto/SystemSettings.kt
```diff
@@ -0,0 +1,9 @@
+package com.epam.drill.dto
+
+import kotlinx.serialization.*
+
+@Serializable
+data class SystemSettings(
+    val packagesPrefixes: List<String> = emptyList(),
+    val sessionIdHeaderName: String = ""
+)
```
viktorbll: newline!

core/src/main/kotlin/com/epam/drill/dto/SystemSettings.kt
```diff
@@ -0,0 +1,9 @@
+package com.epam.drill.dto
```
viktorbll: Who approved this pacakage name?

PR #35, opened 2020-02-11
EPMDJ-2334. Implementing filling nextBuild field of BuildInfo

s0larm00n:


LikhanovMaxim (changes requested):
this build fails with a compile error, rerun checks when merges: "... drill common module: for class BuildInfo field 'nextBuild' is proposed to be added"

viktorbll:
Unapproved incorrect design desicion

PR #38, opened 2020-02-12
EPMDJ-2214 Error finishing all scopes if one is agents is not registered

LikhanovMaxim:
changed a response for dispatch-action


core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
```diff
@@ -274,17 +275,20 @@ class PluginDispatcher(override val kodein: Kodein) : KodeinAware {
         action: String
     ): Pair<HttpStatusCode, Any> {
         val sessionId = UUID.randomUUID().toString()
-        return agents
+        val statusesResponse: MutableList<StatusResponse> = mutableListOf()
```
viktorbll: rewrite with immutable list

viktorbll (approved):

PR #50, opened 2020-03-10
EPMDJ-2416. Implemented agent logging config saving on admin side

s0larm00n:


core/src/main/kotlin/agent/StorableLoggingConfig.kt
```diff
@@ -0,0 +1,25 @@
+package com.epam.drill.admin.agent
+
+import com.epam.drill.api.dto.*
+import com.epam.kodux.*
+import kotlinx.serialization.*
+
+val defaultLoggingConfig = LoggingConfig(
+    warn = true,
+    info = true,
+    debug = true,
```
viktorbll:
```suggestion
    debug = false,
```

core/src/main/kotlin/agent/StorableLoggingConfig.kt
```diff
@@ -0,0 +1,25 @@
...
+    trace = true,
```
viktorbll:
```suggestion
    trace = false,
```

core/src/main/kotlin/agent/StorableLoggingConfig.kt
```diff
@@ -0,0 +1,25 @@
...
+    error = true
+)
+
+@Serializable
+data class StorableLoggingConfig(
+    @Id
+    val agentId: String,
```
viktorbll:
```suggestion
    @Id val agentId: String,
```

core/src/main/kotlin/agent/StorableLoggingConfig.kt
```diff
@@ -0,0 +1,25 @@
...
+    val config: LoggingConfig
+)
+
+fun LoggingConfig.toStorable(agentId: String) = StorableLoggingConfig(
+    agentId = agentId,
+    config = this.copy()
```
viktorbll: An immutable object does not require copying.

PR #51, opened 2020-03-11
EPMDJ-2425 Removed unused endpoints and changed agent config update endpoint

ViktoriiaLebedeva:


core/src/main/kotlin/agent/Dto.kt
```diff
@@ -30,6 +30,15 @@ data class AgentInfoDto(
     val agentType: String
 )
 
+@Serializable
+data class BasicAgentInfoDto(
```
LikhanovMaxim: 1. BasicAgentInfoDto duplicates fields from AgentInfoDto, fix it, for example, by creating in AgentInfoDto a new field with type 'BasicAgentInfoDto'
1. The name 'Basic' is not well for this. I think in a basic agent, we should have 'id', 'serviceGroup', etc instead of random fields. 	It would be better to use Change/Update or smth like this

viktorbll (approved):

PR #61, opened 2020-03-25
Added config (.bat files)

IanaSosiuk:


.gitattributes
```diff
@@ -1 +1,2 @@
 gradlew text eol=lf
+*.bat   text eol=crlf
```
viktorbll, line 2: no newline at the end of file

PR #86, opened 2020-07-17
EPMDJ-2884 Implemented the storing and updating of agent host

ViktoriiaLebedeva:


viktorbll:
Too much code for a simple field

PR #105, opened 2020-09-15
EPMDJ-3613 Fixed case sensitivity for search

ViktoriiaLebedeva:


core/src/main/kotlin/endpoints/plugin/Filters.kt
```diff
@@ -26,7 +26,7 @@ private fun FrontMessage.applyFilter(statement: SearchStatement?): FrontMessage
         @Suppress("UNCHECKED_CAST")
         (this as? Iterable<Any>)?.filter { fMessage ->
             val propertyValue = fMessage::class.getProperty(statement.fieldName).call(fMessage)
-            statement.value in propertyValue.toString()
+            statement.value.toLowerCase() in propertyValue.toString()
```
viktorbll:
```suggestion
            "$propertyValue".contains(statement.value, ignoreCase = true)
```

viktorbll (changes requested):

viktorbll (approved):

PR #110, opened 2020-09-25
EPMDJ-4347 Added tests for bulk operations with notifications

ViktoriiaLebedeva:


core/src/test/kotlin/notifications/NotificationsTest.kt
```diff
@@ -92,4 +92,20 @@ class NotificationsTest {
...
+    fun `updateAll - asc and desc must be updated`() {
+        var notifications = Notifications()
+        val notification = Notification(
+            id = "some-id",
+            agentId = agentId,
+            createdAt = System.currentTimeMillis(),
+            type = NotificationType.BUILD,
+            message = "some-message"
+        )
+        notifications = notifications.plus(notification)
```
viktorbll: "Plus" is an operator - use it as it is supposed to be used.

core/src/test/kotlin/notifications/NotificationsTest.kt
```diff
@@ -92,4 +92,20 @@ class NotificationsTest {
         assertNull(notifications[notification.id])
         assertFalse { notifications.valuesDesc.contains(notification) }
     }
+
+    @Test
+    fun `updateAll - asc and desc must be updated`() {
+        var notifications = Notifications()
```
viktorbll, line 97: The test should be against 2 or more notifications.

viktorbll (changes requested):

core/src/test/kotlin/notifications/NotificationsTest.kt
viktorbll: All tests

viktorbll (changes requested):

viktorbll (approved):

PR #134, opened 2020-12-17
EPMDJ-5724 Fixed: plugin doesn't initialize when you try to add it

ViktoriiaLebedeva:


viktorbll (approved):

core/src/main/kotlin/endpoints/AgentManager.kt
```diff
@@ -296,6 +296,7 @@ class AgentManager(override val kodein: Kodein) : KodeinAware {
                             val pluginId = plugin.pluginBean.id
                             launch {
                                 session.sendPlugin(plugin, updatedInfo).await()
+                                ensurePluginInstance(entry, plugin)
                                 session.enablePlugin(pluginId, agentId).await()
```
viktorbll:
```suggestion
                                ensurePluginInstance(entry, plugin)
                                session.sendPlugin(plugin, updatedInfo).await()
                                session.enablePlugin(pluginId, agentId).await()
```

viktorbll (changes requested):

viktorbll (approved):

PR #148, opened 2021-02-01
EPMDJ-6027 Api change for service group

Mazel-Tovr:


core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
```diff
@@ -98,9 +98,9 @@ internal class PluginDispatcher(override val kodein: Kodein) : KodeinAware {
                             example("")
                         ), notFound()
                     )
-                post<ApiRoot.ServiceGroup.Plugin.DispatchAction, String>(meta) { pluginParent, action ->
+                post<ApiRoot.AgentGroup.Plugin.DispatchAction, String>(meta) { pluginParent, action ->
                     val pluginId = pluginParent.parent.pluginId
-                    val serviceGroupId = pluginParent.parent.parent.serviceGroupId
+                    val serviceGroupId = pluginParent.parent.parent.groupId
```
LikhanovMaxim:
```suggestion
                    val groupId = pluginParent.parent.parent.groupId
```

LikhanovMaxim (changes requested):

core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
Mazel-Tovr: Should I change it everywhere?

LikhanovMaxim (approved):

PR #149, opened 2021-02-03
EPMDJ-5963 [BE] For an Offline Service in a Service group should not start a test session

Mazel-Tovr:


core/src/main/kotlin/endpoints/AgentManager.kt
```diff
@@ -220,6 +220,7 @@ class AgentManager(override val kodein: Kodein) : KodeinAware {
         }.getOrDefault(id, persistentHashMapOf())
         if (instances.isEmpty()) {
             logger.info { "Agent with id '${id}' was disconnected" }
+            entryOrNull(id)?.updateAgent { it.copy(status = AgentStatus.OFFLINE) }?.persistToDatabase()
```
LikhanovMaxim: why here do we save status in db?

tests/src/test/kotlin/e2e/plugin/PluginTest.kt
```diff
@@ -63,7 +63,7 @@ class PluginTest : E2EPluginTest() {
                 println("hi ag3")
             }
             uiWatcher { channel ->
-                waitForMultipleAgents(channel)
+                waitForMultipleAgents(channel, 3)
```
LikhanovMaxim: move to default value

LikhanovMaxim (changes requested):

core/src/main/kotlin/endpoints/AgentManager.kt
Mazel-Tovr: Before the changes, we saved all agent statuses changes in the database, but now, we will store only 2 states in the database, either the agent is not registered, or offline , if you do not save the status to offline, then after restarting the system, the agent will rise with the status not registered. 
Although, maybe it's better to save the agent's status offline immediately after registering the agent ?

tests/src/test/kotlin/e2e/plugin/PluginTest.kt
```diff
@@ -82,11 +83,11 @@ class PluginTest : E2EPluginTest() {
 
     }
 
-    private suspend fun waitForMultipleAgents(channel: Channel<GroupedAgentsDto>) {
+    private suspend fun waitForMultipleAgents(channel: Channel<GroupedAgentsDto>, instanceCount: Int) {
```
LikhanovMaxim:
```suggestion
    private suspend fun waitForMultipleAgents(channel: Channel<GroupedAgentsDto>, instanceCount: Int = 3) {
```

core/src/main/kotlin/endpoints/AgentManager.kt
```diff
@@ -209,6 +209,7 @@ class AgentManager(override val kodein: Kodein) : KodeinAware {
         adminData(agentId).updateSettings(dto.systemSettings)
         pluginsToAdd.forEach { plugin -> ensurePluginInstance(entry, plugin) }
         info.sync()
+        info.copy(status = AgentStatus.OFFLINE).persistToDatabase()
```
LikhanovMaxim: move to another place: in starting admin

Mazel-Tovr:
fixed in https://github.com/Drill4J/admin/pull/153

PR #157, opened 2021-02-25
EPMDJ-6172 General information isn't updated for group after registration

Mazel-Tovr:


core/src/main/kotlin/group/Handler.kt
```diff
@@ -224,7 +224,7 @@ class GroupHandler(override val kodein: Kodein) : KodeinAware {
                 logger.error(e) { "Error registering agent $agentId" }
             }
             async(handler) {
-                agentManager.register(info.id, regInfo.copy(name = agentId, description = agentId))
+                agentManager.register(info.id, regInfo.copy(name = agentId, description = agentId, environment = ""))
```
LikhanovMaxim: What is the value in regInfo.environment, why do we change it? As I see we have a default value of environment = ""

Mazel-Tovr: regInfo. environment-contains the value set for the service group and in the current implementation it will be set for all agents of this group , you are right here you need to put not an empty string, but a string from the agent config

LikhanovMaxim (approved):

PR #159, opened 2021-03-02
EPMDJ-6271 [BE] The "Target Host" field in the "Node.js Agent Settings" is cleared when the "Service Group Settings" are changed

Mazel-Tovr:


core/src/main/kotlin/endpoints/AgentManager.kt
```diff
@@ -489,8 +489,11 @@ class AgentManager(override val kodein: Kodein) : KodeinAware {
             val handler = CoroutineExceptionHandler { _, e ->
                 logger.error(e) { "Error updating agent $agentId" }
             }
+            val settings = systemSettings.takeUnless {
+                info.agentType == AgentType.NODEJS
+            } ?: systemSettings.copy(targetHost = adminData(agentId).settings.targetHost)
```
LikhanovMaxim: I think 'takeUnless' is less readable than takeIf
```suggestion
            val settings = systemSettings.takeIf {
                info.agentType != AgentType.NODEJS
            } ?: systemSettings.copy(targetHost = adminData(agentId).settings.targetHost)
```

Mazel-Tovr: There is another option: we can just write: `systemSettings.copy(targetHost = adminData(agentId).settings.targetHost)`    because we use targetHost - field only in the js agent and it can't affect the settings of other agents @LikhanovMaxim  what do you think about it

LikhanovMaxim (approved):

PR #168, opened 2021-04-01
EPMDJ-6307 [BE] Out of memory when finish session or scope

LikhanovMaxim:
replaced cache impl: use off-heap instead of heap space


core/build.gradle.kts
```diff
@@ -66,7 +69,8 @@ val defaultAppJvmArgs = listOf(
     "-server",
     "-Djava.awt.headless=true",
     "-XX:+UseG1GC",
-    "-XX:+UseStringDeduplication"
+    "-XX:+UseStringDeduplication",
+    "-XX:MaxDirectMemorySize=10G"
```
IgorKey, line 73: why 10Gb?

PR #169, opened 2021-04-02
EPMDJ-6816 Add off-heap cache implementation

LikhanovMaxim:
mapDB; add param 'DRILL_PLUGIN_CACHE_TYPE' to switch impl of cache


core/src/main/kotlin/cache/impl/MapDBCacheService.kt
```diff
@@ -0,0 +1,90 @@
+package com.epam.drill.admin.cache.impl
```
Mazel-Tovr, line 16: add copyright

Mazel-Tovr (approved):

PR #178, opened 2021-04-22
 EPMDJ-6943 Added support for sending files from test2code

Mazel-Tovr:


core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
```diff
@@ -324,6 +324,9 @@ internal class PluginDispatcher(override val kodein: Kodein) : KodeinAware {
             contentType = ContentType.MultiPart.Any,
             status = statusCode
         )
+        is FileResponse -> call.respondFile(
+            file = response.data
```
LikhanovMaxim, line 347: can we remove a temp file after sending it?

Mazel-Tovr: yes, we can not to send file but outputSteam , see new commit

core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
```diff
@@ -357,12 +357,14 @@ internal class PluginDispatcher(override val kodein: Kodein) : KodeinAware {
...
 
+@Suppress("UNCHECKED_CAST")
 private fun Any.toStatusResponse(): WithStatusCode = when (this) {
-    is ActionResult -> when (val d = data) {
-        is String -> StatusMessageResponse(code, d)
-        is File -> FileResponse(code, d)
-        else -> StatusResponse(code, d.actionToJson())
-    }
+    is ActionResult ->
+        when (val d = data) {
```
Mazel-Tovr: `  is ActionResult -> when (val d = data) {`
return as it was

LikhanovMaxim (approved):

LikhanovMaxim (approved):

PR #180, opened 2021-04-28
EPMDJ-6912 Tests are not displayed in the previous build

Mazel-Tovr:


core/src/test/kotlin/cache/impl/MapDBCacheServiceTest.kt
```diff
@@ -131,4 +131,56 @@ class MapDBCacheServiceTest {
...
+                        methodCount = CountDto(covered = 1, total = 98),
+                        count = CountDto(covered = 2, total = 812)),
+                    stats = TestStats(duration = 4894, result = "FAILED")))
+        }
+        cache[key] = data
+        assertEquals(data, cache[key])
+        val cache2: Cache<String, Any> = cacheService.getOrCreate(cacheId, "build2")
+        cache2[key] = ArrayList<Any>()
+        assertEquals(data, cache[key])
+        assertEquals("", cache2[key])
```
LikhanovMaxim: can we do this test simpler?

Mazel-Tovr: yep

LikhanovMaxim (approved):

PR #181, opened 2021-04-29
EPMDJ-6966 [BE] Investigation - how to work with old builds

LikhanovMaxim:
added an action of removing a build data

### DIFF
--- a/core/src/main/kotlin/api/routes/ApiRoot.kt
+++ b/core/src/main/kotlin/api/routes/ApiRoot.kt
@@ -37,8 +37,15 @@ class ApiRoot(val prefix: String = "api") {
     @Location("/version")
     data class Version(val parent: ApiRoot)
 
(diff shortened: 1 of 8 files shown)


Mazel-Tovr (approved):

core/src/main/kotlin/api/routes/ApiRoot.kt
```diff
@@ -37,8 +37,15 @@ class ApiRoot(val prefix: String = "api") {
...
 
-    @Location("/cache-stats")
-    data class CacheStats(val parent: ApiRoot)
+    @Group(SYSTEM)
+    @Location("/cache")
+    data class Cache(val parent: ApiRoot) {
+        @Location("/stats")
+        data class CacheStats(val parent: Cache)
+
+        @Location("/flush")
```
Mazel-Tovr: clear

PR #190, opened 2021-05-26
EPMDJ-7339 Added url decoding in destination

ViktoriiaLebedeva:


LikhanovMaxim (approved):

core/src/main/kotlin/endpoints/WsTopic.kt
```diff
@@ -128,3 +129,7 @@ fun getKSerializer(value: Any): KSerializer<Any> = when (value) {
 } as KSerializer<Any>
 
 fun elementSerializer(collection: Collection<*>) = (collection.firstOrNull() ?: "")::class.serializer()
+
+fun String.urlDecode() = takeIf { '%' in it }?.run {
+    runCatching { URLDecoder.decode(this, Charsets.UTF_8.name()) }.getOrNull()
```
LikhanovMaxim: add logging if we cannot decode url

PR #199, opened 2021-06-10
EPMDJ-7530 add new plugin download support

Mazel-Tovr:


core/src/main/resources/application.conf
```diff
@@ -34,6 +34,8 @@ drill {
         }
     }
     plugins {
+        ids = "test2code,state-watcher"
+        ids = ${?DRILL_PLUGINS_IDS}
         remote {
```
Mazel-Tovr: fixx

PR #206, opened 2021-06-18
EPMDJ-7759 Handle errors when downloading a plugin

Mazel-Tovr:


core/src/test/kotlin/download/PluginDownloadTest.kt
```diff
@@ -0,0 +1,60 @@
...
+        val loaderService = ArtifactoryPluginLoader(Artifactory.GITHUB, pluginStorageDir, false)
+        loaderService.loadPlugins(listOf("unknown plugin"))
+        assert(pluginStorageDir.listFiles()?.size == 0)
+    }
+
+    @Test
+    fun `should successfully download only available plugins`() = runBlocking {
+        val loaderService = ArtifactoryPluginLoader(Artifactory.GITHUB, pluginStorageDir, false)
+        loaderService.loadPlugins(listOf("state-watcher", "unknown plugin"))
+        assert(pluginStorageDir.listFiles()?.size == 1)
```
LikhanovMaxim: Is it ok that tests spend a quota of downloading?

core/src/test/kotlin/download/PluginDownloadTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    @AfterTest
+    fun after() {
+        pluginStorageDir.deleteRecursively()
+    }
+
+    @Test
+    fun `should successfully download t2c plugin`() = runBlocking {
+        val loaderService = ArtifactoryPluginLoader(Artifactory.GITHUB, pluginStorageDir, false)
+        loaderService.loadPlugins(listOf("test2code"))
```
LikhanovMaxim: can you use allowedPlugins or another const?

core/src/test/kotlin/download/PluginDownloadTest.kt
Mazel-Tovr: Well, in our case, when checks very often fall, this is not normal, I think it is worth turning off these tests until we fix the integration tests.

core/src/main/kotlin/plugins/ArtifactoryPluginLoader.kt
```diff
@@ -27,14 +27,19 @@ private val logger = KotlinLogging.logger {}
...
         HttpClient(CIO).use { client ->
             for (pluginId in pluginIds) {
                 val url = pluginId.downloadUrl()
                 if (url.isNotBlank()) {
-                    client.downloadPlugin(url)
+                    runCatching {
+                        client.downloadPlugin(url)
+                    }.onFailure {
+                        logger.warn { "Specified url $url is incorrect, downloading from the specified repository ${artifactory.name}" }
+                        client.load(pluginId)
```
Mazel-Tovr: remove

LikhanovMaxim (changes requested):
one runCatching

core/src/main/kotlin/plugins/PluginLoaderService.kt
```diff
@@ -42,6 +44,7 @@ class PluginLoaderService(
     private val pluginPaths: List<File> = listOf(pluginStoragePath).map { it.canonicalFile }
 
     init {
+
```
ViktoriiaLebedeva: remove it

ViktoriiaLebedeva (changes requested):

LikhanovMaxim (approved):

PR #207, opened 2021-06-24
EPMDJ-7339 Added url decoding in destination

ViktoriiaLebedeva:


core/src/main/kotlin/endpoints/WsTopic.kt
```diff
@@ -128,3 +131,10 @@ fun getKSerializer(value: Any): KSerializer<Any> = when (value) {
 } as KSerializer<Any>
 
 fun elementSerializer(collection: Collection<*>) = (collection.firstOrNull() ?: "")::class.serializer()
+
+fun String.urlDecode() = takeIf { '%' in it }?.run {
+    runCatching {
+        URLDecoder.decode(this, Charsets.UTF_8.name())
+    }.onFailure { logger.error { "Failed to decode string" } }.getOrNull()
+} ?: this
+
```
Mazel-Tovr: it might be better to move this method into Util.kt file ?

ViktoriiaLebedeva: There is only one method in Util.kt that is not used. In my opinion, Util.kt should be removed. We only decode ws destinations, so I added it to the WsTopic.kt file.

core/src/main/kotlin/endpoints/WsTopic.kt
```diff
@@ -128,3 +131,9 @@ fun getKSerializer(value: Any): KSerializer<Any> = when (value) {
 } as KSerializer<Any>
 
 fun elementSerializer(collection: Collection<*>) = (collection.firstOrNull() ?: "")::class.serializer()
+
+fun String.urlDecode() = takeIf { '%' in it }?.run {
+    runCatching {
+        URLDecoder.decode(this, Charsets.UTF_8.name())
+    }.onFailure { logger.error { "Failed to decode string" } }.getOrNull()
```
LikhanovMaxim: Can we add the string to the log which cannot decode?

PR #208, opened 2021-06-28
EPMDJ-7785 Async dispatch action for microservices

ViktoriiaLebedeva:


LikhanovMaxim (commented):
need to check with scratch

PR #212, opened 2021-07-02
EPMDJ-7812 Fixed: information about methods isn't displayed on "all builds" page

ViktoriiaLebedeva:


core/src/main/kotlin/plugin/Cache.kt
```diff
@@ -57,13 +57,12 @@ class PluginCaches(
...
         buildVersion: String,
         type: String,
-    ): Any? = plugins.keys.firstOrNull()?.let { pluginId ->
-        retrieveMessage(
-            pluginId,
-            AgentSubscription(agentId, buildVersion),
-            "/data/$type"
-        )
-    }.takeIf { it != "" }
+        pluginId: String = "test2code"
```
LikhanovMaxim: const
tech debt

PR #232, opened 2021-11-12
EPMDJ-8765 Add ability to send messages from admin part of plugin

ViktoriiaLebedeva:


core/src/main/kotlin/plugin/Senders.kt
```diff
@@ -78,6 +82,20 @@ class PluginSenders(override val kodein: Kodein) : KodeinAware {
...
+        override suspend fun sendToAgent(agentId: String, message: Any) {
+            message.actionSerializerOrNull()?.let { serializer ->
+                val actionStr = serializer stringify message
+                val agentAction = PluginAction(pluginId, actionStr)
+                agentManager.agentSessions(agentId).map {
+                    //TODO EPMDJ-8233 move to the api
+                    it.sendToTopic<Communication.Plugin.DispatchEvent, PluginAction>(
+                        agentAction,
+                        topicName = "/plugin/action/${actionStr.encodeBase64()}"
+                    )
```
LikhanovMaxim: check await

PR #239, opened 2021-12-01
EPMDJ-6500 Incorrect encoding of test type in Cyrillic

Mazel-Tovr:


core/src/main/kotlin/config/Content.kt
```diff
@@ -22,12 +22,36 @@ import io.ktor.request.*
...
+                charset = call.request.contentCharset()
+                    ?: Charsets.UTF_8
+            )
+            else -> null
+        }
+        if (result != null)
+            proceedWith(ApplicationReceiveRequest(query.typeInfo, result, true))
+    }
+}
+
```
Mazel-Tovr, line 54: just change header

Mazel-Tovr: header are immutable (((

PR #249, opened 2022-01-19
EPMDJ-9704 [BE] Error when finishing non-empty active scopes in the S…

LikhanovMaxim:
…ervice Group


core/src/main/resources/application.conf
```diff
@@ -61,7 +61,7 @@ drill {
         dbName = ${?DRILL_DB_USER_NAME}
         password = "mysecretpassword"
         password = ${?DRILL_DB_PASSWORD}
-        maximumPoolSize = 3
+        maximumPoolSize = 10
         maximumPoolSize = ${?DRILL_DB_MAX_POOL_SIZE}
```
LikhanovMaxim, line 65: postgres optimal connection pool size

core/src/main/kotlin/DrillApplication.kt
```diff
@@ -135,6 +135,7 @@ fun Application.module() = kodeinApplication(
             this.maximumPoolSize = maxPoolSize
             this.isAutoCommit = false
             this.transactionIsolation = "TRANSACTION_REPEATABLE_READ"
+            this.leakDetectionThreshold = 15 * 1000L
```
LikhanovMaxim: remove

PR #251, opened 2022-01-24
EPMDJ-8630 Add ability to remove not registered and preregistered agents

Mazel-Tovr:


core/src/main/kotlin/endpoints/AgentManager.kt
```diff
@@ -108,6 +108,22 @@ class AgentManager(override val kodein: Kodein) : KodeinAware {
         } else null
     }
 
+    //TODO EPMDJ-9872 Add ability to remove offline agent
+    internal suspend fun removeAgent(
+        agentId: String,
+    ): Boolean = (get(agentId) ?: loadAgentInfo(agentId)).let { storedInfo ->
+        val agentStatus = storedInfo?.agentStatus
+        if (storedInfo == null || agentStatus == AgentStatus.NOT_REGISTERED || agentStatus == AgentStatus.PREREGISTERED) {
```
LikhanovMaxim, line 116: storedInfo  == null
what is the case?

core/src/main/kotlin/endpoints/admin/AgentEndpoints.kt
```diff
@@ -203,6 +203,25 @@ class AgentEndpoints(override val kodein: Kodein) : KodeinAware {
...
+                val meta = "Remove all agent info"
+                    .responds(
+                        ok<Unit>(), notFound(), badRequest()
+                    )
+                delete<ApiRoot.Agents.Agent>(meta) { payload ->
+                    val agentId = payload.agentId
+                    if (agentManager.removeAgent(agentId)) {
+                        call.respond(HttpStatusCode.OK, EmptyContent)
+                    } else {
+                        logger.debug { "Deleting registered Agent '$agentId' isn't supported yet" }
```
LikhanovMaxim, line 218: add TODO EPMDJ-9872

core/src/main/kotlin/endpoints/AgentManager.kt
Mazel-Tovr: This is for the case when we don't have AgentInfo, but there is a PreparedAgentData.

PR #256, opened 2022-02-03
EPMDJ-9860 Web Analytics Implementation

Mazel-Tovr:


core/src/main/kotlin/version/Dto.kt
```diff
@@ -38,7 +39,16 @@ data class AdminVersionDto(
     val java: String = "",
 )
 
+@Serializable
+data class AnalyticsInfoDto(
+    val clientId: String,
```
Mazel-Tovr, line 44:
```suggestion
    val clientId: String,
    val analyticDisable : Boolean,
```

core/src/main/kotlin/api/routes/WsRoot.kt
```diff
@@ -32,4 +32,7 @@ object WsRoot {
 
     @Location("/api/groups/{groupId}")
     class Group(val groupId: String)
+
+    @Location("/api/analytics")
```
Mazel-Tovr:
```suggestion
    @Location("/api/analytics/info")
```

analytics/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
+plugins {
```
Mazel-Tovr, line 1: Add read.ME

PR #266, opened 2022-03-22
EPMDJ-9811 [BE] Finish scope when agent is offline

LikhanovMaxim:
added ability to invoke dispatch actions for offline agents

### DIFF
--- a/core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
+++ b/core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
@@ -16,7 +16,8 @@
 package com.epam.drill.admin.endpoints.plugin
 
 import com.epam.drill.admin.agent.*
-import com.epam.drill.admin.api.agent.*
+import com.epam.drill.admin.api.agent.AgentStatus.*
+import com.epam.drill.admin.api.agent.BuildStatus.*
 import com.epam.drill.admin.api.plugin.*
 import com.epam.drill.admin.api.routes.*
 import com.epam.drill.admin.api.websocket.*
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
```diff
@@ -120,17 +121,20 @@ internal class PluginDispatcher(override val di: DI) : DIAware {
                     val (statusCode, response) = agent?.run {
                         val plugin: Plugin? = this@PluginDispatcher.plugins[pluginId]
                         if (plugin != null) {
-                            if (info.agentStatus == AgentStatus.REGISTERED && buildManager.buildStatus(agentId) == BuildStatus.ONLINE) {
+                            val buildStatus = buildManager.buildStatus(agentId)
+                            if (isReadyToAction(agentId) || (action.contains("SWITCH_ACTIVE_SCOPE") && buildStatus != BUSY)) {
+                                //todo EPMDJ-10291 remove hardcode SWITCH_ACTIVE_SCOPE
```
LikhanovMaxim: add action to check is it possible to do smth for offline agent

core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
```diff
@@ -120,17 +121,32 @@ internal class PluginDispatcher(override val di: DI) : DIAware {
...
-                            if (info.agentStatus == AgentStatus.REGISTERED && buildManager.buildStatus(agentId) == BuildStatus.ONLINE) {
+                            val buildStatus = buildManager.buildStatus(agentId)
+                            //todo refactor if it is OK
+                            val isActionForOffline = this[pluginId]?.let { adminPart ->
+                                val action2 = """
+                                    {
+                                      "type": "IS_POSSIBLE_OFFLINE",
+                                      "payload": $action
+                                    }
+                                """.trimIndent()
```
Mazel-Tovr: brr , it's so weird. can we do stm like this 
```
    @Serializable
    data class Action(
        val type: String = "IS_POSSIBLE_OFFLINE",
        val payload: String,
    )

    Action.serilizer() stringify Action(payload = payload)
```

PR #278, opened 2022-04-12
EPMDJ-10291 [BE][Admin] Remove additional action from admin

Mazel-Tovr:


core/src/main/kotlin/endpoints/plugin/PluginDispatcher.kt
```diff
@@ -348,7 +343,7 @@ internal class PluginDispatcher(override val di: DI) : DIAware {
     }
 
     private fun Agent.isReadyToAction(agentId: String = info.id) = info.agentStatus == REGISTERED
-            && buildManager.buildStatus(agentId) == ONLINE
+            && buildManager.buildStatus(agentId) != BUSY
```
Mazel-Tovr: remove busy check

PR #302, opened 2023-06-27
Feature/transfer class metadata epmdj 10603

iryabov:


test-framework/src/main/kotlin/com/epam/drill/e2e/plugin/loadPlugin.kt
```diff
@@ -115,6 +115,14 @@ suspend fun AdminTest.loadPlugin(
                 memoryClassLoader.loadClass(clsName)
             }
         }
+
+        if (spykAgentPart is ClassScanner) {
```
RomanDavlyatshin, line 119: This change allows to remove `get-load-classes-datas` calls in all e2e tests, right?

iryabov: Yes. This is a universal approach for any tests based on a test framework.

PR #304, opened 2023-08-01
Feature/migrate to logging framework epmdj 10617

epamrsa:


admin-core/build.gradle.kts
```diff
@@ -62,7 +62,6 @@ dependencies {
 
     implementation(project(":admin-analytics"))
     implementation(project(":common"))
-    implementation(project(":logger"))
```
RomanDavlyatshin, line 65: `admin` has no own logger configuration?

RomanDavlyatshin: admin has own logging implementation using the same framework

RomanDavlyatshin (approved):

PR #305, opened 2023-08-03
Feature/embedded test2code epmdj 10620

epamrsa:


RomanDavlyatshin (changes requested):
change frametype handler in `admin-core/src/main/kotlin/com/epam/drill/admin/endpoints/agent/AgentHandler.kt` to log warning when unknown frame type is received

PR #315, opened 2023-09-18
feature: rework adding coverage at CoverDataPart handler

bodyangug:


test2code-admin/src/main/kotlin/com/epam/drill/plugins/test2code/Plugin.kt
```diff
@@ -363,16 +363,20 @@ class Plugin(
...
+                    if (activeScope.activeSessions[sessionId] == null){
+                        sessionId = GLOBAL_SESSION_ID
+                    }
+                    activeScope.activeSessions[sessionId].let{
                         activeScope.addProbes(sessionId) { data }?.run {
                             if (isRealtime) {
                                 activeScope.probesChanged()
                             }
                         }
-                    } ?: logger.debug { "Attempting to add coverage in non-existent active session with id $sessionId" }
```
RomanDavlyatshin: This check and a logger call are still required, since there might be no global session in `activeScope.activeSessions` list (In case it might've been finished, but not started yet)

RomanDavlyatshin: We should think about _always_ keeping global session alive (in current backend implementation).
E.g. we can add check to `finishSession` handler to look if a just-finished session were a global one, and if it was - start a new one.
Hacky - yes, but it might work just fine

test2code-admin/src/main/kotlin/com/epam/drill/plugins/test2code/Plugin.kt
```diff
@@ -363,16 +363,20 @@ class Plugin(
...
         is CoverDataPart -> {
-            message.data.groupBy { it.sessionId }
+            message.data
+                .groupBy { it.sessionId }
                 .forEach { (probeSessionId, data) ->
-                    val sessionId = probeSessionId ?: message.sessionId ?: GLOBAL_SESSION_ID
-                    activeScope.activeSessions[sessionId]?.let {
+                    var sessionId = probeSessionId ?: message.sessionId ?: GLOBAL_SESSION_ID
+                    if (activeScope.activeSessions[sessionId] == null){
+                        sessionId = GLOBAL_SESSION_ID
```
RomanDavlyatshin: Code is a bit contrived, since `sessionId` might already contain `GLOBAL_SESSION_ID` (so re-assigning it won't help at all)
Lets think of a better solution

RomanDavlyatshin (changes requested):

test2code-admin/src/main/kotlin/com/epam/drill/plugins/test2code/Plugin.kt
bodyangug: log message was returned

test2code-admin/src/main/kotlin/com/epam/drill/plugins/test2code/Plugin.kt
bodyangug: let`s impl new logic at new release

RomanDavlyatshin (approved):
Looks good to me
@iryabov please look at this as well, just to be sure everything is ok

PR #319, opened 2023-10-03
feat: add Basic Auth for rest services

iryabov:
EPMDJ-10664

### DIFF
--- a/admin-core/src/main/kotlin/com/epam/drill/admin/endpoints/admin/AgentEndpoints.kt
+++ b/admin-core/src/main/kotlin/com/epam/drill/admin/endpoints/admin/AgentEndpoints.kt
@@ -48,7 +48,7 @@ class AgentEndpoints(override val di: DI) : DIAware {
     init {
         app.routing {
 
-            authenticate {
+            authenticate("jwt", "basic") {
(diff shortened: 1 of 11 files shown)


admin-core/src/main/kotlin/com/epam/drill/admin/endpoints/admin/AgentEndpoints.kt
```diff
@@ -48,7 +48,7 @@ class AgentEndpoints(override val di: DI) : DIAware {
     init {
         app.routing {
 
-            authenticate {
+            authenticate("jwt", "basic") {
```
RomanDavlyatshin, line 51: Move respective routes under the single `authenticate() {}` block to avoid duplication
It should work: https://ktor.io/docs/authentication.html#authenticate-route

RomanDavlyatshin (changes requested):

admin-core/src/main/kotlin/com/epam/drill/admin/endpoints/admin/AgentEndpoints.kt
iryabov: Fixed

PR #321, opened 2023-10-03
refactor: remove isGlobal property

bodyangug:


RomanDavlyatshin:
obsoleted by move to new http transport

PR #339, opened 2023-11-08
POC implementation of oauth2 using Keycloak

iryabov:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+package com.epam.drill.admin.auth.config
+
+import io.ktor.application.*
+import org.kodein.di.DI
+import org.kodein.di.DIAware
+import org.kodein.di.instance
+
+class OAuthConfig(override val di: DI) : DIAware {
+    private val app by instance<Application>()
```
RomanDavlyatshin, line 24: purpose (class does nothing atm)

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
```diff
@@ -54,25 +70,119 @@ class SecurityConfig(override val di: DI) : DIAware {
...
         jwt(name) {
             realm = "Access to the http(s) and the ws(s) services"
-            verifier(jwtTokenService.verifier)
+            verifier((tokenService as JwtTokenService).verifier)
             validate {
                 it.payload.toPrincipal()
             }
             authHeader { call ->
-                val headerValue = call.request.headers[HttpHeaders.Authorization] ?: "Bearer ${call.parameters["token"]}"
+                val headerValue = call.request.headers[Authorization] ?: "Bearer ${call.request.cookies["jwt"] ?: call.parameters["token"]}"
```
RomanDavlyatshin: _tbd_ reason for 3 branches to figure out "header" value:
- "legacy" header auth?
- `Bearer ...` cookie for new implementation?
- `&token=...` param for ws connection?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
```diff
@@ -54,25 +70,119 @@ class SecurityConfig(override val di: DI) : DIAware {
...
+                httpClient.post<HttpResponse>("http://localhost:8080/realms/master/protocol/openid-connect/logout") {
+                    headers {
+                        append(HttpHeaders.Authorization, "Bearer ${principal.accessToken}")
+                    }
+                }
+            }
+        }
+        get("/oauth/callback") {
+            val principal: OAuthAccessTokenResponse.OAuth2? = call.principal()
+            if (principal != null) {
```
RomanDavlyatshin: _Style nitpick_: do not use `else` alternate branch to handle exceptions.

Prefer to handle exceptional scenarios first (flatter code, easier to understand):
```
if (checkErrorCondition()) { 
  // throw Error // call.unauthorizedError() // whatever required to process error
}

// rest of the code
getUserInfo(...)
call.response...
...
```

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
```diff
@@ -54,25 +70,119 @@ class SecurityConfig(override val di: DI) : DIAware {
...
+                getUserInfo(principal.accessToken)
+
+                call.response.cookies.append(Cookie("jwt", principal.accessToken, httpOnly = true, path = "/"))
+                call.respondRedirect("/")
```
RomanDavlyatshin: _future todo_: configurable redirect target

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
```diff
@@ -54,25 +70,119 @@ class SecurityConfig(override val di: DI) : DIAware {
...
+
+private suspend fun getUserInfo(accessToken: String): User? {
+    val httpClient = HttpClient(Apache)
+    val response =
+        httpClient.get<HttpResponse>("http://localhost:8080/realms/master/protocol/openid-connect/userinfo") {
+            headers {
+                append(Authorization, "Bearer $accessToken")
+            }
+        }
+    return response.status.takeIf { it.isSuccess() }?.let { _ ->
```
RomanDavlyatshin, line 158: missing negative scenario handling? ( when `response.status.isSuccess()` returns `false`)

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
```diff
@@ -54,25 +70,119 @@ class SecurityConfig(override val di: DI) : DIAware {
...
+                    headers {
+                        append(HttpHeaders.Authorization, "Bearer ${principal.accessToken}")
+                    }
+                }
+            }
+        }
+        get("/oauth/callback") {
+            val principal: OAuthAccessTokenResponse.OAuth2? = call.principal()
+            if (principal != null) {
+                getUserInfo(principal.accessToken)
```
RomanDavlyatshin: `getUserInfo` output is unused

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
```diff
@@ -54,25 +70,119 @@ class SecurityConfig(override val di: DI) : DIAware {
...
+            }
+        }
+    return response.status.takeIf { it.isSuccess() }?.let { _ ->
+        val stringBody = response.receive<String>()
+        val jsonBody = Json.parseToJsonElement(stringBody)
+        jsonBody.toPrincipal()
+    }
+}
+
+private fun JsonElement.toPrincipal(): User {
```
RomanDavlyatshin, line 165: _Style nitpick_: IMO pushing this code out to the extension function makes it more confusing (especially bc it is extending JsonElement)

Lambda could be employed to make code more readable at the call site:
```
fun getUserInfo() {
...
        val stringBody = response.receive<String>()
        val jsonBody = Json.parseToJsonElement(stringBody)

        val toPrinciple = { jsonBody: JsonBody -> 
          User(
            username = jsonObject["username"]?.jsonPrimitive?.contentOrNull ?: "",
            role = findRole(jsonObject["roles"]?.jsonArray?.map { it.toString() } ?: emptyList())
          )
        }
       return toPrinciple(jsonBody)
}
```  
or just leave it inlined
```
fun getUserInfo() {
        //... 
        val stringBody = response.receive<String>()
        val jsonObject = Json.parseToJsonElement(stringBody).jsonObject

        return User(
          username = jsonObject["username"]?.jsonPrimitive?.contentOrNull ?: "",
          role = findRole(jsonObject["roles"]?.jsonArray?.map { it.toString() } ?: emptyList())
        )
}
```

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
```diff
@@ -54,25 +70,119 @@ class SecurityConfig(override val di: DI) : DIAware {
...
+    val jsonObject = this.jsonObject
+    return User(
+        username = jsonObject["username"]?.jsonPrimitive?.contentOrNull ?: "",
+        role = findRole(jsonObject["roles"]?.jsonArray?.map { it.toString() } ?: emptyList())
```
RomanDavlyatshin: initialization details (such as defaults for fields) can be move to `User`'s constructor:
- improves readability
- no need to duplicate those checks in other places, where `User()` might be instantiated

Will boil down to
```
  User (
    userName = jsonObject["username"]?.jsonPrimitive
    role = jsonObject["roles"]?.jsonArray?.map { it.toString() }?.let { findRole(it) }
    // pushing findRole to the end makes comprenehsion easier since code written order matches code execution order
  )
```

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/route/RouteUtils.kt
```diff
@@ -33,10 +33,10 @@ suspend fun ApplicationCall.validationError(cause: Exception) {
     respond(HttpStatusCode.BadRequest, MessageResponse(cause.message ?: "User data is invalid"))
 }
 
-suspend fun ApplicationCall.unauthorizedError(cause: Exception) {
-    respond(HttpStatusCode.Unauthorized, MessageResponse(cause.message ?: "User is not authenticated"))
+suspend fun ApplicationCall.unauthorizedError(cause: Exception? = null) {
```
RomanDavlyatshin, line 36: -Use case for nullable `Exception`
-Call parameter just `(exception: Exception)` since `cause` is actually a field inside `Exception`  - adds a little bit of confusion

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/JwkTokenService.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    override fun issueToken(user: UserView): String {
+        throw UnsupportedOperationException("Issuing tokens is not supported")
+    }
+
+    override fun verifyToken(token: String) {
+        getVerifier(provider, issuer, token).verify(token)
+    }
+
+    private fun getVerifier(
```
RomanDavlyatshin, line 53: _tbd_ reason for implementing it so specifically, instead of just supplying JWK like here [1](https://auth0.com/blog/adding-auth0-authorization-to-a-ktor-http-api/) or here [2](https://ktor.kotlincn.net/servers/features/authentication/jwt.html)

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/JwkTokenService.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.epam.drill.admin.auth.config.OAuthConfig
+import com.epam.drill.admin.auth.model.UserView
+import com.epam.drill.admin.auth.service.TokenService
+import io.ktor.auth.jwt.*
+import java.net.URL
+import java.security.interfaces.ECPublicKey
+import java.security.interfaces.RSAPublicKey
+import java.util.concurrent.TimeUnit
+
+class JwkTokenService(oauthConfig: OAuthConfig) : TokenService {
```
RomanDavlyatshin, line 36: `JwkTokenService` -> `OauthJwkTokenService`?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
```diff
@@ -54,25 +70,119 @@ class SecurityConfig(override val di: DI) : DIAware {
...
         username = subject,
         role = Role.valueOf(getClaim("role").asString())
     )
 }
 
+private fun findRole(roleNames: List<String>): Role {
+    return Role.values().find { role ->
+        roleNames.toSet().map { it.lowercase() }.contains(role.name.lowercase())
+    } ?: Role.UNDEFINED
+}
```
RomanDavlyatshin, line 184: is this some sort of the "mapping", that matches the first match from "external" roles to "our" roles?

admin-core/src/main/kotlin/com/epam/drill/admin/core/WsExtension.kt
```diff
@@ -55,9 +55,9 @@ fun Route.authWebSocket(
 }
 
 private suspend fun DefaultWebSocketServerSession.socketAuthentication(tokenService: TokenService) {
-    val token = call.parameters["token"]
+    val token = call.request.cookies["jwt"] ?: call.parameters["token"]
```
RomanDavlyatshin, line 58: I suspect there is an option in WebSocket clients to pass token other than using parameter
_TODO_: investigate and remove this `parameters["token"]` option if true

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
RomanDavlyatshin: will be extended l8ter

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
RomanDavlyatshin: yes

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
RomanDavlyatshin: refactored in the next prs

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
RomanDavlyatshin: or java context path / ktor analogue

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
RomanDavlyatshin: will be used when sessions impl is ready

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
RomanDavlyatshin: will be refactored later due to other reason

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/route/RouteUtils.kt
RomanDavlyatshin: `ApplicationCall.unauthorizedError()` is handy

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
iryabov: We will need to think about whether all 3 are worth supporting

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
iryabov: added TODO

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
iryabov: Fixed

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
iryabov: In this case, this function will return null, this value will be handled.

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
iryabov: I've removed the call for now

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
iryabov: Done

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/JwkTokenService.kt
iryabov: This code will be refactored

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/JwkTokenService.kt
iryabov: This code will be refactored. The entire class will likely be removed.

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SecurityConfig.kt
iryabov: Yes. But this mapping will be improved in the non-POC version

admin-core/src/main/kotlin/com/epam/drill/admin/core/WsExtension.kt
iryabov: We will need to choose one way to pass the token to the WS connection. An attempt is made here to maintain backward compatibility, because... we didn't discuss the opposite

PR #342, opened 2023-11-15
feat: add dockerfile to build image for windows platform

bodyangug:


admin-core/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN mkdir distr
+RUN mkdir work
+
+COPY build/libs/admin-core-0.8.3-all.jar ./app/admin-core-0.8.3-all.jar
+
+EXPOSE 8090
+EXPOSE 5006
+EXPOSE 8453
+
+CMD ["java", "-server", "-Djava.awt.headless=true", "-XX:+UseG1GC","-XX:+UseStringDeduplication", "-XX:MaxDirectMemorySize=2G", "-cp", "/app/admin-core-0.8.3-all.jar", "io.ktor.server.netty.EngineMain"]
```
RomanDavlyatshin: Hardcoded `.jar` name won't work - it'll be updated in next versions
We either have to:
- fix name to something unchangeable `drill-admin.jar`
_OR_
- pass version via ARG at build time

admin-core/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM eclipse-temurin:11.0.21_9-jdk-nanoserver
+
+RUN mkdir distr
+RUN mkdir work
+
+COPY build/libs/admin-core-0.8.3-all.jar ./app/admin-core-0.8.3-all.jar
```
RomanDavlyatshin: see comment on harcoded jar name below

admin-core/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM eclipse-temurin:11.0.21_9-jdk-nanoserver
+
+RUN mkdir distr
```
RomanDavlyatshin: Each docker command creates a separate docker layer (with its own cache) - avoid spamming RUNs -> prefer to use `&&` if commands belong to the same logical "step":
```
RUN  mkdir && \
mkdir work
```
`&& \` - is equivalent to `&&`. Backslash `\` escapes line return symbol, so visually separate lines will  be processed as a single line

admin-core/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN mkdir work
+
+COPY build/libs/admin-core-0.8.3-all.jar ./app/admin-core-0.8.3-all.jar
+
+EXPOSE 8090
```
RomanDavlyatshin: Allow to customize port values via env variable (make sure to set defaults)
```
# check exact syntax with docs / examples - I'm writing this by memory
ENV MY_PORT=8090
EXPOSE $MY_PORT
```

I suggest to name those API_PORT, DEBUG_PORT, SECURE_API_PORT respectively

admin-core/build.gradle.kts
```diff
@@ -192,6 +192,14 @@ tasks {
     withType<JibTask> {
         dependsOn(processJibExtraDirs)
     }
+    val createDockerImage by registering(Exec::class) {
+        workingDir(projectDir)
+        commandLine(
+            "docker", "build",
+            "-t", "ghcr.io/drill4j/admin:$version",
```
RomanDavlyatshin: Move string with the registry/org/image names to a variable above (somewhere next to `var jvmArgs...`) and reuse it here and in `jib {` section
```
var fullImageTag = "ghcr.io/drill4j/admin"
...
"-t", "$fullImageTag:$version"

```

RomanDavlyatshin (changes requested):

admin-core/Dockerfile
bodyangug: chanhed

admin-core/Dockerfile
bodyangug: done

admin-core/Dockerfile
bodyangug: done

admin-core/Dockerfile
bodyangug: done

admin-core/build.gradle.kts
bodyangug: done

admin-core/build.gradle.kts
```diff
@@ -192,6 +202,33 @@ tasks {
     withType<JibTask> {
         dependsOn(processJibExtraDirs)
     }
+    val loginToDocker by registering(Exec::class) {
+        dependsOn(assemble)
+        workingDir(projectDir)
+        commandLine("docker", "login", "ghcr.io", "-u $gitUsername", "-p $gitPassword")
```
RomanDavlyatshin: move `"ghcr.io"` to `val` above and use it both here and in `fullImageTag` (to avoid duplication and possible issues with we swap registry in one place and forget in the other)

RomanDavlyatshin (changes requested):

.github/workflows/release.yml
```diff
@@ -98,13 +99,14 @@ jobs:
...
+            ${{ (matrix.config.preset == 'mingwX64') && 'assemble loginToDocker createWindowsDockerImage pushWindowsDockerImage' || '' }}
         env:
-          GRADLE_OPTS: >
-            -Djib.to.image=ghcr.io/drill4j/admin
-            -Djib.to.auth.username=${{ secrets.GH_PACKAGE_USERNAME }}
-            -Djib.to.auth.password=${{ secrets.GH_PACKAGE_TOKEN }}
+          GH_USERNAME: ${{ secrets.GH_PACKAGE_USERNAME }}
+          GH_TOKEN: ${{ secrets.GH_PACKAGE_TOKEN }}
       - uses: actions/upload-release-asset@v1
+#        if: matrix.config.preset == 'linuxX64'
```
RomanDavlyatshin: either remove or uncomment

.github/workflows/release.yml
```diff
@@ -113,6 +115,7 @@ jobs:
           asset_name: admin-${{ needs.release.outputs.version }}.zip
           asset_content_type: application/zip
       - uses: actions/upload-release-asset@v1
+#        if: matrix.config.preset == 'linuxX64'
```
RomanDavlyatshin: either remove or uncomment

bodyangug: uncommented

.github/workflows/release.yml
bodyangug: uncommented

admin-core/build.gradle.kts
bodyangug: moved

admin-core/build.gradle.kts
```diff
@@ -122,24 +122,41 @@ application {
...
+        auth {
+            username=gitUsername
+            password=gitPassword
+        }
     }
     container {
-        ports = listOf("8090", "5006", "8453")
-        volumes = listOf("/distr", "/work")
+        ports = listOf(apiPort, debugPort , secureApiPort)
+        volumes = listOf("/config", "/config/ssl")
```
RomanDavlyatshin, line 146: I would like us to have a _single_ volume - to reduce complexity
E.g. in Kubernetes creating volumes requires 3 steps, 1 config for each-> 2 volumes is x2 config lines

Everything can be placed in one volume:
```
- application.conf
- somesslcert.jks
- something-else.foo
```

admin-core/build.gradle.kts
```diff
@@ -122,24 +122,41 @@ application {
...
         mainClass = jarMainClassName
         jvmFlags = defaultJvmArgs
     }
     extraDirectories {
-        setPaths(jibExtraDirs)
-        permissions = mapOf("/distr" to "775", "/work" to "775")
+        setPaths("/config/ssl")
```
RomanDavlyatshin, line 151: _tbd_ explain purpose

admin-core/build.gradle.kts
```diff
@@ -122,24 +122,41 @@ application {
...
+        permissions = mapOf("/config" to "775","/config/ssl" to "775")
+        paths {
+            path{
+                setFrom("build/resources/main/")
```
RomanDavlyatshin, line 155: lines 155 - 157 just do the copying of the `application.conf` - right?

admin-core/src/main/resources/application.conf
```diff
@@ -2,18 +2,16 @@ ktor {
     deployment {
         port = 8090
```
RomanDavlyatshin, line 3: I don't see how we can disable this port via configuration.
Lets think of something

admin-core/src/main/resources/application.conf
```diff
@@ -2,18 +2,16 @@ ktor {
...
         modules = [com.epam.drill.admin.DrillApplicationKt.module]
     }
     security {
         ssl {
-            keyStore = temporary.jks
-            keyAlias = drill-admin
-            keyStorePassword = changeit
-            privateKeyPassword = changeit
+            keyStore = ${?SSL_KEY_STORE_PATH}
+            keyAlias = ${?SSL_KEY_ALIAS}
```
RomanDavlyatshin, line 12: Inconsistent naming

RomanDavlyatshin (changes requested):

admin-core/build.gradle.kts
bodyangug: right

admin-core/build.gradle.kts
bodyangug: ssl folder will be removed

admin-core/src/main/resources/application.conf
bodyangug: That value will be override by external config at comand block

admin-core/build.gradle.kts
RomanDavlyatshin: on the second though, having two volumes instead of one is better, since we can swap files inidivudally
we'll keep it like that

admin-core/src/main/resources/application.conf
bodyangug: renamed

RomanDavlyatshin (approved):

PR #346, opened 2023-11-21
Storing OAuth users in database

iryabov:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/entity/UserEntity.kt
```diff
@@ -18,8 +18,8 @@ package com.epam.drill.admin.auth.entity
 data class UserEntity(
     var id: Int? = null,
     var username: String,
-    var passwordHash: String,
-    var role: String,
+    var passwordHash: String? = null,
+    var role: String? = null,
```
iryabov: The result of the /userinfo OAuth2 request is mapped to the existing UserEntity class, and not to a new class, to avoid unnecessary mapping between entities, but this had to be remove not nullable from the role field

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+class OAuthServiceImpl(
+    private val httpClient: HttpClient,
+    private val oauthConfig: OAuthConfig,
+    private val userRepository: UserRepository
+) : OAuthService {
+
+    override suspend fun signInThroughOAuth(principal: OAuthAccessTokenResponse.OAuth2): UserInfoView {
+        val oauthUser = getUserInfo(principal.accessToken).toEntity()
+        val dbUser = userRepository.findByUsername(oauthUser.username)
+        if (dbUser?.blocked == true)
```
iryabov, line 43: I didn't add a check for undefined role to return 403 on login like I did it in simple auth, because throwing an exception will cause a rollback of the transaction and we will not save the user in the database

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+            }
+            .receive<String>()
+            .let { Json.parseToJsonElement(it) }
+    }.onFailure { cause ->
+        throw OAuthUnauthorizedException("User info request failed: ${cause.message}", cause)
+    }.getOrThrow()
+}
+
+private fun JsonElement.toEntity(): UserEntity = UserEntity(
+    username = jsonObject.getValue("preferred_username").jsonPrimitive.content,
```
iryabov, line 73: Will be refactored in configurable user mapping issue

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthModule.kt
```diff
@@ -43,13 +38,19 @@ const val JWT_COOKIE = "jwt"
 
 
 val oauthDIModule = DI.Module("oauth") {
-    configureOAuthDI()
+    configureSimpleAuthDI()
```
RomanDavlyatshin: Reason for adding simple auth's DI to oauth DI configuration

admin-core/src/main/kotlin/com/epam/drill/admin/DrillApplication.kt
```diff
@@ -113,6 +121,21 @@ fun Application.moduleWithOAuth2() {
     routing {
         drillAdminRoutes()
         configureOAuthRoutes()
+        route("/api") {
+            authenticate("jwt") {
+                userInfoRoute()
```
iryabov, line 127: I didn't add sign-in, sign-up, update-password, reset-password routes because they contradict external authentication way

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthModule.kt
```diff
@@ -43,13 +38,19 @@ const val JWT_COOKIE = "jwt"
...
     configureJwtDI()
+    configureOAuthDI()
 }
 
 fun DI.Builder.configureOAuthDI() {
     bind<HttpClient>("oauthHttpClient") with singleton { HttpClient(Apache) }
-    bind<OAuthConfig>() with singleton { OAuthConfig(di) }
+    bind<OAuthConfig>() with singleton { OAuthConfig(instance<Application>().environment.config) }
+    bind<OAuthService>() with singleton { TransactionalOAuthService(OAuthServiceImpl(
+        httpClient = instance("oauthHttpClient"),
```
RomanDavlyatshin, line 51: passing `"oauthHttpClient"` string literal in place of type / class / class instance seems odd

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
RomanDavlyatshin: _tbd_

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/entity/UserEntity.kt
RomanDavlyatshin: _tbd_ I see no reason to make `role` nullable - can't we fallback to `Role.UNDEFINED` in all cases?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/PasswordServiceImpl.kt
```diff
@@ -28,7 +28,8 @@ class PasswordServiceImpl(
         return BCrypt.hashpw(password, BCrypt.gensalt())
     }
 
-    override fun matchPasswords(candidate: String, hashed: String): Boolean {
+    override fun matchPasswords(candidate: String, hashed: String?): Boolean {
+        if (hashed == null) return false
```
RomanDavlyatshin: _tbd_
1. I'm not sure this is the correct approach. Returning `false` here will result in "invalid credentials" type of message
2. Where as we, probably, should indicate that user must login using different authentication mechanism
   - or we shouldn't, depending on if we want to expose _how_ are users are authenticated

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+import io.ktor.http.*
+import kotlinx.serialization.json.*
+
+class OAuthServiceImpl(
+    private val httpClient: HttpClient,
+    private val oauthConfig: OAuthConfig,
+    private val userRepository: UserRepository
+) : OAuthService {
+
+    override suspend fun signInThroughOAuth(principal: OAuthAccessTokenResponse.OAuth2): UserInfoView {
```
RomanDavlyatshin, line 40: Method name seems contrived - simple `signIn` would do?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+        val oauthUser = getUserInfo(principal.accessToken).toEntity()
+        val dbUser = userRepository.findByUsername(oauthUser.username)
+        if (dbUser?.blocked == true)
+            throw OAuthUnauthorizedException("User is blocked")
```
RomanDavlyatshin: 1. Strictly speaking, this is not "`OAuth`" type of error, since users are blocked in _Drill4J_'s instance, and not on oauth provider side
2. This info may be too specific? I mean 401 with default message would work just fine

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    private suspend fun createOrUpdateUser(
+        oauthUser: UserEntity,
+        dbUser: UserEntity?
+    ): UserEntity = dbUser
+        ?.merge(oauthUser)
+        ?.apply { userRepository.update(this) }
+        ?: oauthUser
+            .run { userRepository.create(this) }
+            .let { oauthUser.copy(id = it) }
```
RomanDavlyatshin: _tbd_ `.let { oauthUser.copy(id = it) }` seems hack-ish

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+
+private fun JsonElement.toEntity(): UserEntity = UserEntity(
+    username = jsonObject.getValue("preferred_username").jsonPrimitive.content,
+    role = jsonObject["roles"]
+        ?.jsonArray
+        ?.map { it.jsonPrimitive.content }
+        .let { findRole(it)?.name }
+)
+
+private fun findRole(roleNames: List<String>?): Role? = roleNames
```
RomanDavlyatshin, line 81: I would rather call it `mapRole`

RomanDavlyatshin: + add fallback to `Role.UNDEFINED` (after discussion)

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+private fun UserEntity.merge(other: UserEntity) = copy(
+    role = other.role ?: this.role
+)
+
+private fun UserEntity.toView(): UserInfoView {
+    return UserInfoView(
+        username = this.username,
+        role = Role.valueOf(this.role ?: Role.UNDEFINED.name)
```
RomanDavlyatshin: Oh, so there is a fallback to `Role.UNDEFINED`
lets discuss it

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -0,0 +1,100 @@
...
+    ?.distinct()
+    ?.map { it.lowercase() }
+    ?.let { roleNamesList ->
+        Role.values().find { role ->
+            roleNamesList.contains(role.name.lowercase())
+        }
+    }
+
+private fun UserEntity.merge(other: UserEntity) = copy(
+    role = other.role ?: this.role
```
RomanDavlyatshin: 1. It isn't really "merge" - its more of a role reassignment.
   If we see no other fields to merge, we'd rename it or, better yet, inline it

2. We're juggling `role` in 3 functions (only in this file + 2 files below)  - this seems very contrived

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/UserManagementServiceImpl.kt
```diff
@@ -84,7 +84,7 @@ private fun UserEntity.toView(): UserView {
     return UserView(
         id = this.id,
         username = this.username,
-        role = Role.valueOf(this.role),
+        role = Role.valueOf(this.role ?: Role.UNDEFINED.name),
```
RomanDavlyatshin: _tbd_ Role default fallback in multiple places

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/UserAuthenticationServiceImpl.kt
```diff
@@ -73,6 +73,6 @@ private fun RegistrationPayload.toUserEntity(passwordHash: String): UserEntity {
 private fun UserEntity.toView(): UserInfoView {
     return UserInfoView(
         username = this.username,
-        role = Role.valueOf(this.role)
+        role = Role.valueOf(this.role ?: Role.UNDEFINED.name)
```
RomanDavlyatshin, line 76: _tbd_ Role default fallback in multiple places

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthModuleTest.kt
```diff
@@ -184,6 +191,70 @@ class OAuthModuleTest {
...
+            }
+        }) {
+            with(handleRequest(HttpMethod.Get, "/oauth/callback?code=$wrongAuthenticationCode&state=$testState")) {
+                assertEquals(HttpStatusCode.Unauthorized, response.status())
+            }
+        }
+    }
+
+    @Test
+    fun `given OAuthUnauthorizedException, oauth callback request must fail with 401 status`() {
```
RomanDavlyatshin: Isn't that test too internals-specific? _tbd_ In which cases `OAuthUnauthorizedException` is thrown? Can't we trigger those events, instead of directly throwing it?

_if not_
rename to `if signInThroughOAuth throws OAuthUnauthorizedException, oauth callback request must fail with 401 status` - `given` seems odd here, since exception is not an "input" per se

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthModuleTest.kt
```diff
@@ -192,19 +263,15 @@ class OAuthModuleTest {
...
-            HttpClient(MockEngine { request ->
-                requestHandlers
-                    .find { request.url.encodedPath == it.first }
-                    ?.runCatching { this@MockEngine.second(request) }
-                    ?.getOrElse { exception ->
-                        respondError(HttpStatusCode.BadRequest, exception.message ?: "${exception::class} error")
-                    }
-                    ?: respondBadRequest()
-            })
+            mockHttpClient(*requestHandlers)
```
RomanDavlyatshin, line 266: might as well inline `mockHttpClient`, or is it called elsewhere?

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        val httpClient = mockHttpClient(
+            "/userInfoUrl" to userInfoResponse(testAccessToken, testUsername, testRole)
+        )
+        val oauthService = OAuthServiceImpl(httpClient, OAuthConfig(config), userRepository)
+        whenever(userRepository.findByUsername(testUsername)).thenReturn(null)
+        whenever(userRepository.create(any())).thenReturn(1)
+
+        val userInfo = oauthService.signInThroughOAuth(withPrincipal(testAccessToken))
+        verify(userRepository).create(any())
+        assertEquals(testUsername, userInfo.username)
```
RomanDavlyatshin: Strictly speaking, none of the asserts actually check for entity existence
I think we should either spy on `userRepository.create` call, or supply actual repository and query it

- Asserting correct username might be ok
- Role looks like another scenario, since it can be changed by mapping

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        val oauthService = OAuthServiceImpl(httpClient, OAuthConfig(config), userRepository)
+        whenever(userRepository.findByUsername(testUsername)).thenReturn(UserEntity(
+            id = 1,
+            username = testUsername,
+            role = Role.USER.name
+        ))
+
+        val userInfo = oauthService.signInThroughOAuth(withPrincipal(testAccessToken))
+        verify(userRepository).update(any())
+        assertEquals(testUsername, userInfo.username)
```
RomanDavlyatshin, line 93: assert of username seems out of place
I would suggest the same approach as with prior test - spy/stub `userRepository.update` call and assert passed params

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthModuleTest.kt
RomanDavlyatshin: https://github.com/Drill4J/admin/pull/346/files#diff-7b08c14a24bb5583550550d9bab8b28b8758f7bf05f28df2d01255d9727cbb1cR114

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        assertEquals(Role.USER, userInfo.role)
+    }
+
+    @Test
+    fun `given blocked OAuth2 principal, signInThroughOAuth must failed`(): Unit = runBlocking {
```
RomanDavlyatshin: `must failed` -> `must fail`

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        put("drill.auth.oauth2.userInfoUrl", "http://some-oauth-server.com/userInfoUrl")
+    }
+
+    @BeforeTest
+    fun setup() {
+        MockitoAnnotations.openMocks(this)
+    }
+
+    @Test
+    fun `given new OAuth2 principal, signInThroughOAuth must create new user`() = runBlocking {
```
RomanDavlyatshin: `given ... OAuth2 principal` _tbd_

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+        ))
+
+        val principal = withPrincipal(testAccessToken)
+        assertThrows<OAuthUnauthorizedException> {
+            oauthService.signInThroughOAuth(principal)
+        }
+    }
+
+    @Test
+    fun `given invalid accessToken, signInThroughOAuth must failed`(): Unit = runBlocking {
```
RomanDavlyatshin: must failed -> must fail

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    private fun withPrincipal(testAccessToken: String) = OAuthAccessTokenResponse.OAuth2(
+        accessToken = testAccessToken,
+        tokenType = "Bearer",
+        expiresIn = 3600,
+        refreshToken = null
+    )
+
+    private fun userInfoResponse(
```
RomanDavlyatshin: rename to `assertUserInfoResponseSuccess`

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/TestUtils.kt
```diff
@@ -93,4 +96,14 @@ suspend fun HttpRequestData.formData() = String(this.body.toByteArray())
 fun URL.queryParams() = query?.split("&")?.associate {
     val (key, value) = it.split("=")
     key to value
-} ?: emptyMap()
\ No newline at end of file
+} ?: emptyMap()
+
+fun mockHttpClient(vararg requestHandlers: RequestHandler) = HttpClient(MockEngine { request ->
+    requestHandlers
+        .find { request.url.encodedPath == it.first }
```
RomanDavlyatshin: Using pair's first/second fields always introduces some level of confusion
I suggest to use data class instead, so "first" and "second" would have actual names

admin-core/src/main/kotlin/com/epam/drill/admin/DrillApplication.kt
```diff
@@ -169,6 +184,14 @@ private fun Application.installPlugins() {
     install(RoleBasedAuthorization)
 }
 
+private fun StatusPages.Configuration.defaultStatusPages() {
+    exception<Throwable> { cause ->
+        logger.error(cause) { "Build application finished with exception" }
```
RomanDavlyatshin: I don't get what `"Build application"` means
If I understand correctly, this log is always capturing request handling exception. If that is correct, I suggest to write `"Failed to process the request"` and make sure to add both request details and exception details to trace (request url, payload, headers, cookies, etc + exception message and stacktrace)

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthModule.kt
RomanDavlyatshin: _tbd_ for a future change? more like "commonDI" ? YAGNI?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthModule.kt
RomanDavlyatshin: "oauthHttpClient" is used in multiple places, hence DI is still handy

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/PasswordServiceImpl.kt
RomanDavlyatshin: - push `if (hashed == null)` check out of this function. `matchPasswords` should not be executed for oauth user
- I don't like identifying "oauth" users by implicit passwordHash==null check  - it seems unreliable and contrived

RomanDavlyatshin: think about more

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
RomanDavlyatshin: We could keep it for clarity, specially if instance context mssing / `this` is implicit

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
iryabov: Changed the exception and the status code to 403 forbidden

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/PasswordServiceImpl.kt
iryabov: - I think the password matching service works with null hashes as expected and does not take on unnecessary responsibility. 
- Сhecking for null outside the matching service will be a little more difficult than just checking for null, because the passwordHash field has a var (not a val) definition
- I don’t think that in the simple authentication service we need to explicitly check for the provider through which the user was registered, we just need to make sure that we won’t be able to log in with any password

iryabov: In any case, I simplified the implementation a little and maybe in this form it raises fewer questions:
    
```
    override fun matchPasswords(candidate: String, hashed: String?): Boolean {
        return hashed != null && BCrypt.checkpw(candidate, hashed)
    }
```

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
iryabov: 1. We are still in the process of OAuth2 authenticating there 
2. I changed to 403 status and remove this message for user

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
iryabov: Ok, I will rename this function in the user mapping branch

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthModuleTest.kt
iryabov: This event is triggered when there are errors in retrieving or processing user info request. But this part of process is a mock here.

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthModuleTest.kt
iryabov: Yes, three times

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/PasswordServiceImpl.kt
RomanDavlyatshin: > In any case, I simplified the implementation a little and maybe in this form it raises fewer questions:
> 
> ```
>     override fun matchPasswords(candidate: String, hashed: String?): Boolean {
>         return hashed != null && BCrypt.checkpw(candidate, hashed)
>     }
> ```

I would say it raises pretty much the same questions 😄 But I agree with your points above. Resolving that conversation

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
RomanDavlyatshin: > We are still in the process of OAuth2 authenticating there

Yes, but my concern is that it'll look confusing outside of this function (e.g. in log).
"Hey look, `OAuthUnauthorizedException` happened, because user is blocked" - there is ambiguity in _where_ user is blocked, and it will require digging in Drill4J code to find out where the issue is.

We can keep the exception type, just make it clear (somehow) that user is blocked in Drill4J db and not by OAuth provider 

> ... and remove this message for user

_tbd_

RomanDavlyatshin: I guess dedicated exception type resolves the issue + we can document that
https://github.com/Drill4J/admin/pull/346/commits/839dd6ee1381c2fb11086549da2572fa2d2d0a1e#diff-588f6a4c58b26ece3d7d18ca210f0598213569fc05c7f673f62e8b073bf41cd5R118

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthModule.kt
```diff
@@ -109,8 +109,13 @@ fun Routing.configureOAuthRoutes() {
...
 
+class OAuthAccessDeniedException(message: String? = null) : RuntimeException(message)
+
 fun StatusPages.Configuration.oauthStatusPages() {
     exception<OAuthUnauthorizedException> { cause ->
-        call.respond(HttpStatusCode.Unauthorized, cause.message ?: "User is not authenticated through OAuth2 provider")
+        call.respond(HttpStatusCode.Unauthorized, cause.message ?: "Failed to verify authentication through OAuth2 provider")
+    }
+    exception<OAuthAccessDeniedException> { cause ->
+        call.respond(HttpStatusCode.Forbidden, cause.message ?: "Access denied")
```
RomanDavlyatshin, line 120: We should document (for end-users, drill4j instance administrators) all these authentication flow scenarios / behaviors when we're done with auth implementation

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -52,39 +50,42 @@ class OAuthServiceTest {
...
+                "/userInfoUrl" shouldRespond assertTokenAndRespondSuccess(testAccessToken, testUsername, testRole)
+            )
+            val oauthService = OAuthServiceImpl(httpClient, OAuthConfig(config), userRepository)
+            whenever(userRepository.findByUsername(testUsername)).thenReturn(null)
+            whenever(userRepository.create(any())).thenReturn(1)
+
+            val userInfo = oauthService.signInThroughOAuth(withPrincipal(testAccessToken))
+            verify(userRepository).create(any())
+            assertEquals(testUsername, userInfo.username)
+            assertTrue(testRole.equals(userInfo.role.name, true))
```
RomanDavlyatshin, line 70: `assertEquals(testRole.name, userInfo.role.name)`?

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -52,39 +50,42 @@ class OAuthServiceTest {
...
+        whenever(userRepository.findByUsername(testUsername)).thenReturn(
+            UserEntity(
+                id = 1,
+                username = testUsername,
+                role = Role.USER.name
+            )
+        )
 
         val userInfo = oauthService.signInThroughOAuth(withPrincipal(testAccessToken))
         verify(userRepository).update(any())
```
RomanDavlyatshin, line 92: I still see no reason to assert username in _this_ scenario

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
```diff
@@ -52,39 +50,42 @@ class OAuthServiceTest {
...
+
+            val userInfo = oauthService.signInThroughOAuth(withPrincipal(testAccessToken))
+            verify(userRepository).create(any())
+            assertEquals(testUsername, userInfo.username)
+            assertTrue(testRole.equals(userInfo.role.name, true))
+        }
 
     @Test
-    fun `given OAuth2 principal that was already logged in, signInThroughOAuth must update user role`() = runBlocking {
+    fun `given user that is authenticated through OAuth2 again, signInThroughOAuth must update user`() = runBlocking {
```
RomanDavlyatshin, line 74: 1. `given existing user, authenticating with OAuth2, signInThroughOAuth must update user entity with data from "user-info"`
2. _tbd_ splitting test case

admin-core/src/main/kotlin/com/epam/drill/admin/DrillApplication.kt
```diff
@@ -186,7 +187,7 @@ private fun Application.installPlugins() {
 
 private fun StatusPages.Configuration.defaultStatusPages() {
     exception<Throwable> { cause ->
-        logger.error(cause) { "Build application finished with exception" }
+        logger.error(cause) { "Failed to process the request ${this.context.request.path()}" }
```
RomanDavlyatshin, line 190: Can we write more request data to log (`request.path` will only tell us _which_ request handler failed, but not the input parameters - headers, body, cookies, etc)

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
RomanDavlyatshin: return the whole entity from `.create`

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
RomanDavlyatshin: will be addressed in the next pr

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
RomanDavlyatshin: this case is split up in the next prs

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
RomanDavlyatshin: this case is split up in the next prs

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
RomanDavlyatshin: `, true)` - enable case-insensitive string comparison

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
RomanDavlyatshin: only change name

admin-core/src/main/kotlin/com/epam/drill/admin/DrillApplication.kt
RomanDavlyatshin: 1. Logging body by default is a bad idea. It may contain sensitive data

2. It is best to enable such verbose logging only when necessary - there is a handy ktor lib for that
TODO create task to add ktor-logger
https://github.com/Koriit/ktor-logging

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthServiceTest.kt
iryabov: Will be done in user mapping branch

PR #349, opened 2023-11-23
Feature/prevent self destructing epmdj 10711

iryabov:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/UserAuthenticationServiceImpl.kt
```diff
@@ -72,6 +72,7 @@ private fun RegistrationPayload.toUserEntity(passwordHash: String): UserEntity {
 
 private fun UserEntity.toView(): UserInfoView {
     return UserInfoView(
+        id = this.id!!, //the user entity must be inserted and have id
```
RomanDavlyatshin, line 75: Avoid using null-check op since it throws bare NPE without any context
I suggest to either handle it, or better yet add check before constructing `UserInfoView()` to explicitly state that `this.id` must not be null

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
```diff
@@ -74,6 +74,7 @@ class OAuthServiceImpl(
 }
 
 private fun UserEntity.toView() = UserInfoView(
```
RomanDavlyatshin, line 76: we do have almost the same function at `UserAuthenticationServiceImpl.kt`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/route/UserAuthenticationRoutes.kt
```diff
@@ -66,6 +66,10 @@ fun StatusPages.Configuration.simpleAuthStatusPages() {
         logger.trace(cause) { "403 Access denied" }
         call.accessDeniedError(cause)
     }
+    exception<SelfDestructException> { cause ->
```
RomanDavlyatshin: 1. I would rather call it something less extreme 😄 
e.g. `ForbiddenOperation` with error `"Cannot modify own profile"`
2. we intend to show this error on UI, so that must have a clear message - no need to include HTTP status code in the message itself

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/UserAuthenticationServiceImpl.kt
RomanDavlyatshin: already address in the next pr

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
RomanDavlyatshin: move extension function to `UserInfoView`

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthServiceImpl.kt
iryabov: Will be done in the PR #351

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/route/UserAuthenticationRoutes.kt
iryabov: 1, Changed to ForbiddenOperationException
2. This is just a log entry and is not shown to users.

PR #351, opened 2023-11-24
Feature/user registration date epmdj 10712

iryabov:
For discussion:
- I decided to make the registration date a nullable field because I believe that the default value is "0" for dates worse than null and not only this
- I decided to set the registration date to current date inside the DatabaseUserRepository otherwise it would not be possible to test with the mocked UserRepository.

### DIFF
--- a/admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/UserAuthenticationServiceImpl.kt
+++ b/admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/UserAuthenticationServiceImpl.kt
@@ -66,14 +66,14 @@ private fun RegistrationPayload.toUserEntity(passwordHash: String): UserEntity {
     return UserEntity(
         username = this.username,
         passwordHash = passwordHash,
-        role = Role.UNDEFINED.name,
+        role = Role.UNDEFINED.name
     )
 }
 
 private fun UserEntity.toView(): UserInfoView {
     return UserInfoView(
-        id = this.id!!, //the user entity must be inserted and have id
+        id = this.id ?: throw NullPointerException("User must be inserted and have id"),
         username = this.username,
-        role = Role.valueOf(this.role ?: Role.UNDEFINED.name)
+        role = Role.valueOf(this.role)
     )
 }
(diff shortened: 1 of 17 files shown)


admin-auth/src/test/kotlin/com/epam/drill/admin/auth/UserRepositoryImplTest.kt
```diff
@@ -79,9 +84,31 @@ class UserRepositoryImplTest {
             assertEquals(userEntity.passwordHash, it[UserTable.passwordHash])
             assertEquals(userEntity.role, it[UserTable.role])
             assertEquals(userEntity.blocked, it[UserTable.blocked])
+            assertEquals(userEntity.registrationDate, it[UserTable.registrationDate])
         }
     }
 
+    @Test
+    fun `given not specified registration date, create must insert user and return registration date equal to current time`() =
```
RomanDavlyatshin: I don't really like the approach since it introduces indeterminism
Time functions are commonly abstracted (with either DI framework or just a parameter of function/class field) - it makes it easy to mock them and make tests repeatable
_tbd_

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/UserManagementTests.kt
```diff
@@ -101,14 +108,18 @@ class UserManagementTest {
                 assertEquals(HttpStatusCode.OK, response.status())
                 val response: UserView = assertResponseNotNull(UserView.serializer())
                 assertEquals("admin", response.username)
+                assertEquals(Role.ADMIN, response.role)
+                assertEquals(testRegistrationDate.toKotlinLocalDateTime(), response.registrationDate)
```
RomanDavlyatshin: is asserting time really necessary in _this_ test case?

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/UserRepositoryImplTest.kt
RomanDavlyatshin: `DatabaseUserRepository(dateNowProvider = {LocalDate.date.now()})`

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/UserAuthenticationServiceImpl.kt
```diff
@@ -66,14 +66,14 @@ private fun RegistrationPayload.toUserEntity(passwordHash: String): UserEntity {
     return UserEntity(
         username = this.username,
         passwordHash = passwordHash,
-        role = Role.UNDEFINED.name,
+        role = Role.UNDEFINED.name
     )
 }
 
 private fun UserEntity.toView(): UserInfoView {
```
iryabov: move extension function to UserInfoView

PR #352, opened 2023-11-27
Feature/simple auth with cookie

iryabov:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/route/UserAuthenticationRoutes.kt
```diff
@@ -91,6 +92,7 @@ fun Route.signInRoute() {
         val userView = authService.signIn(loginPayload)
         val token = tokenService.issueToken(userView)
         call.response.header(HttpHeaders.Authorization, token)
```
RomanDavlyatshin: Passing token in _both_ header and http-only cookie defeats the purpose of httpOnly cookie mechanism - to make token inaccessible via javascript in the webbrowser.
Lets discuss if we can keep it _only_ in cookie. I don't see much difference in extracting value from header of from cookie in terms of usability

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/route/UserAuthenticationRoutes.kt
RomanDavlyatshin: TODO create task in Jira

PR #353, opened 2023-11-30
Feature/javadocs for auth epmdj 10692

iryabov:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
```diff
@@ -31,40 +34,82 @@ class OAuthConfig(private val config: ApplicationConfig) {
...
+    /**
+     * UI application root path. The key is "drill.ui.rootPath". Optional, "/" by default.
+     */
     val uiRootPath: String
         get() = ui.propertyOrNull("rootPath")?.getString() ?: "/"
 
+    /**
+     * OAuth2 access token payload mapping.
+     *
+     * Used for map access token claims to [com.epam.drill.admin.auth.model.UserInfoView].
```
RomanDavlyatshin: `Used for map` sounds grammatically incorrect 🤔  -> change to either `Used for mapping` or `Used to map`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
```diff
@@ -31,40 +34,82 @@ class OAuthConfig(private val config: ApplicationConfig) {
...
+     * Includes username and roles mapping:
+     * A username mapping is a name of field containing the username field in the access token payload. Optional, "sub" by default.
```
RomanDavlyatshin: 1. Very mouthful, hard to understand 😄 
Lets change to:
```
"username" must contain name of a claim. Value of that claim is going to be treated as username.
"roles" must contain name of a claim which value contains list of roles. Leave empty for no mapping.
NOTE: this is only the _name_ of the claim containing roles - actual mapping is specified in ["roleMapping"] section
```

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
```diff
@@ -31,40 +34,82 @@ class OAuthConfig(private val config: ApplicationConfig) {
...
         get() = oauth2.config("tokenMapping").run {
             UserMapping(
-                username = propertyOrNull("username")?.getString() ?: "sub", //see https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
+                username = propertyOrNull("username")?.getString() ?: "sub",
                 roles = propertyOrNull("roles")?.getString()
             )
         }
 
+    /**
+     * OAuth2 user info mapping.
```
RomanDavlyatshin, line 107: State precedence over token payload, e.g.:
`If /user-info mapping is specified values extracted from token [tokenMapping] are ignored (values mapped from /user-info take precedence)`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
```diff
@@ -31,40 +34,82 @@ class OAuthConfig(private val config: ApplicationConfig) {
...
+     *
+     * Used for map response to [com.epam.drill.admin.auth.model.UserInfoView].
+     * Includes username and roles mapping:
+     * A username mapping is a name of field containing the username field in the OAuth2 user info response. Optional, "username" by default.
```
RomanDavlyatshin: -> 
```
"username" must contain name of a field in /user-info response body. Value of that field is going to be treated as username.`
"roles" must contain name of a field in /user-info response body response which value contains list of roles. Leave empty for no mapping.
NOTE: this is only the _name_ of the field containing roles - actual mapping is specified in ["roleMapping"] section
```

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
```diff
@@ -73,6 +118,14 @@ class OAuthConfig(private val config: ApplicationConfig) {
             )
         }
 
+    /**
+     * A mapping OAuth2 roles to Drill4J roles.
```
RomanDavlyatshin: 1. Remove leading `"A "` (it sounds grammatically incorrect in this particular case) -> `Mapping configuration from OAuth2 roles to Drill4J roles.`
2. What if multiple OAuth2 roles map to the same Drill4J role (e.g. "devops", "lead" and "product manager" must map to "Role.ADMIN", and "developer", "manual qa" must map to "Role.USER"

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
RomanDavlyatshin: Explain that `userInfoUrl` is the key that defines

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuthConfig.kt
RomanDavlyatshin: 2 -> separate task

PR #355, opened 2023-12-04
Feature/simple auth with oauth2 epmdj 10728

iryabov:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/AuthConfig.kt
```diff
@@ -1,10 +1,26 @@
...
+ */
 package com.epam.drill.admin.auth.config
 
 import io.ktor.config.*
 
 enum class AuthType {
     SIMPLE,
-    OAUTH2
+    OAUTH2,
+    SIMPLE_WITH_OAUTH2
```
RomanDavlyatshin: `SIMPLE_WITH_OAUTH2` -> `SIMPLE_AND_OAUTH2`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuth2Module.kt
```diff
@@ -111,6 +98,15 @@ fun Authentication.Configuration.configureOAuthAuthentication(di: DI) {
     }
 }
 
+fun Authentication.Configuration.configureBasicStubAuthentication() {
+    basic("basic") {
+        realm = "Access to the http(s) services"
+        validate {
+            null //Basic authentication is not supported for the OAuth2 provider, but must be declared
```
RomanDavlyatshin: _tbd_ why

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SimpleAuthModule.kt
```diff
@@ -194,6 +189,11 @@ private val Application.userRepoType: UserRepoType
         ?.getString()?.let { UserRepoType.valueOf(it) }
         ?: UserRepoType.DB
 
+private fun isExternalRoleManagement(config: OAuth2Config?): Boolean {
+    return (config?.userInfoUrl == null && config?.tokenMapping?.roles != null) ||
```
RomanDavlyatshin, line 190: 1. Extract to named `val`s for clarity
```
val fromTokenContent = config?.userInfoUrl == null && config?.tokenMapping?.roles != null
val fromUserInfoResponse = config?.userInfoUrl != null && config.userInfoMapping.roles != null
return fromTokenContent || fromUserInfoResponse
```
2. This code implies that you cannot pull user data from `/user-info`, but map roles from token content. Are we sure about that?

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuth2Module.kt
RomanDavlyatshin: add comment that it is required as a stub for ktor configuration

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/SimpleAuthModule.kt
RomanDavlyatshin: 2 - yes This is the simplification we are willing to accept

PR #356, opened 2023-12-05
Feature/create api for api keys epmdj 10730

iryabov:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/entity/ApiKeyEntity.kt
```diff
@@ -0,0 +1,12 @@
+package com.epam.drill.admin.auth.entity
+
+import java.time.LocalDateTime
+
+data class ApiKeyEntity(
+    val id: Int? = null,
+    val user: UserEntity,
```
RomanDavlyatshin, line 7: motivation for keeping the whole entity, rather than the user id

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/entity/ApiKeyEntity.kt
```diff
@@ -0,0 +1,12 @@
...
+    val description: String,
+    val apiKeyHash: String,
+    val expired: LocalDateTime,
```
RomanDavlyatshin: `expired` -> `expiresAt`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/model/ApiKeyModels.kt
```diff
@@ -0,0 +1,36 @@
...
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class ApiKeyView(
+    val id: Int,
+    val userId: Int,
+    val description: String,
+    val expired: LocalDateTime,
+    val created: LocalDateTime,
+    val username: String,
```
RomanDavlyatshin, line 14: R we sure about having both `userId` and `username`?

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/UserApiKeyTests.kt
```diff
@@ -0,0 +1,117 @@
...
+                getUserApiKeysRoute()
+            }
+        }) {
+            with(handleRequest(HttpMethod.Get, "/user-keys") {
+                addHeader(HttpHeaders.ContentType, ContentType.Application.Json.toString())
+                addJwtToken(username = "test-user", userId = testUserId)
+            }) {
+                assertEquals(HttpStatusCode.OK, response.status())
+                val response: List<ApiKeyView> = assertResponseNotNull(ListSerializer(ApiKeyView.serializer()))
+                assertEquals(2, response.size)
```
RomanDavlyatshin, line 51: I don't see the part where we populate repository width data or mock it. Am I missing something?

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/entity/ApiKeyEntity.kt
RomanDavlyatshin: resolution: we won't store the whole entity, but rather the specific fields

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/entity/ApiKeyEntity.kt
RomanDavlyatshin: `created` -> `createdAt`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/model/ApiKeyModels.kt
RomanDavlyatshin: we need userId to allow navigating/performing user-related operations

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/UserApiKeyTests.kt
RomanDavlyatshin: not implemented yet

iryabov:
Should be discussed:
I decided that the `ApiKeyView` for management purposes should contain information about the username, user role, etc., but the `UserApiKeyView` should not because they are already in the context of the current user.
So I added a nullable user field to the ApiKeyEntity. For management purposes, `api_key` table will be joined with the `user` table in the database, but in case of user purposes it will not be. This is for SQL optimization.

PR #360, opened 2023-12-15
Feature/auth modules refactoring epmdj 10742

iryabov:
Refactor OAuth2 modules, so that you can regulate which authentication is enabled/disabled through the enabled settings

### DIFF
--- a/admin-core/src/main/kotlin/com/epam/drill/admin/DrillApplication.kt
+++ b/admin-core/src/main/kotlin/com/epam/drill/admin/DrillApplication.kt
@@ -46,7 +46,6 @@ import io.ktor.routing.*
 import io.ktor.websocket.*
 import mu.*
 import org.flywaydb.core.*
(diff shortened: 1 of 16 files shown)


admin-core/src/main/kotlin/com/epam/drill/admin/DrillApplication.kt
```diff
@@ -56,119 +55,37 @@ private val logger = KotlinLogging.logger {}
...
-        }
-    }
-}
 
-@Suppress("unused")
-fun Application.moduleWithOAuth2() {
     install(StatusPages) {
-        simpleAuthStatusPages()
-        oauthStatusPages()
+        simpleAuthStatusPages() //TODO rename so as not to be associated with simple form based auth
```
RomanDavlyatshin: just "authStatusPages" ?

RomanDavlyatshin (approved):

PR #364, opened 2023-12-18
Feature/api key cleanup epmdj 10746

iryabov:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/JwtConfig.kt
```diff
@@ -53,7 +53,7 @@ class JwtConfig(private val config: ApplicationConfig) {
      * A lifetime of a JWT. Optional, 30 minutes by default.
```
RomanDavlyatshin: Comment became outdated
Just leave it w/o mentioning exact value (might be misleading, easily becomes obsolete)

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/model/AuthConfigModels.kt
```diff
@@ -0,0 +1,41 @@
...
+@Serializable
+data class AuthConfigView(
+    val simpleAuth: SimpleAuthConfigView? = null,
+    val oauth2: OAuth2ConfigView? = null
+)
+
+@Serializable
+data class SimpleAuthConfigView(
+    val enabled: Boolean,
+    val signUpEnabled: Boolean
```
RomanDavlyatshin, line 17: Can we have simple auth but disable signUp?
One case I see: migration from simple auth to oauth2 and external users, but we want to keep "old" users intact

admin-core/src/main/resources/application.conf
```diff
@@ -98,20 +92,12 @@ drill {
             roleMapping {
             }
         }
-        simpleAuth {
+        simple {
             enabled = true
             enabled = ${?DRILL_SIMPLE_AUTH_ENABLED}
```
RomanDavlyatshin: env variable name became inconsistent
should it be DRILL_AUTH_SIMPLE_ENABLED ?

admin-core/src/main/resources/application.conf
```diff
@@ -98,20 +92,12 @@ drill {
...
-        password {
-            minLength = 6
-            minLength = ${?DRILL_PASSWORD_MIN_LENGTH}
-            mustContainUppercase = false
-            mustContainUppercase = ${?DRILL_PASSWORD_CONTAIN_UPPERCASE}
-            mustContainLowercase = false
-            mustContainLowercase = ${?DRILL_PASSWORD_CONTAIN_LOWERCASE}
-            mustContainDigit = false
-            mustContainDigit = ${?DRILL_PASSWORD_CONTAIN_DIGIT}
+            passwordStrength {
```
RomanDavlyatshin, line 99: Reason for removing defaults? It will require us/user to dig through module implementation to set these. I suggest to set all params to "false"/unset values, but keep the setting names

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/JwtConfig.kt
RomanDavlyatshin: leave value for javadoc (if we add it in the future) just update to the right number

admin-core/src/main/resources/application.conf
RomanDavlyatshin: We already don't have all settings here -> this is _not_ a single source of truth for all settings
Add exhaustive configuration description to website docs

PR #367, opened 2023-12-19
feat: add oauth2 error handling to html page

iryabov:
EPMDJ-10748

### DIFF
--- a/admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuth2Module.kt
+++ b/admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuth2Module.kt
@@ -25,10 +25,12 @@ import io.ktor.application.*
 import io.ktor.auth.*
 import io.ktor.client.*
 import io.ktor.client.engine.apache.*
+import io.ktor.content.*
 import io.ktor.features.*
 import io.ktor.http.*
 import io.ktor.response.*
 import io.ktor.routing.*
+import mu.KotlinLogging
 import org.kodein.di.DI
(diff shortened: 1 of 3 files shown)


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuth2Module.kt
```diff
@@ -135,9 +149,38 @@ class OAuthAccessDeniedException(message: String? = null) : RuntimeException(mes
...
     exception<OAuthUnauthorizedException> { cause ->
-        call.respond(HttpStatusCode.Unauthorized, cause.message ?: "Failed to verify authentication through OAuth2 provider")
+        logger.error("Failed to verify authentication through OAuth2 provider", cause)
+        call.respond(
+            htmlContent("Unauthorized Error", HttpStatusCode.Unauthorized)
+        )
     }
     exception<OAuthAccessDeniedException> { cause ->
-        call.respond(HttpStatusCode.Forbidden, cause.message ?: "Access denied")
+        logger.error("Failed to verify authentication through OAuth2 provider", cause)
```
RomanDavlyatshin: make error log distinct from OAuthUnauthorizedException

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/OAuth2Module.kt
RomanDavlyatshin: add statuses

PR #368, opened 2023-12-20
feat: add the ability to specify oauth2 role mapping as a comma separ…

iryabov:
…ated list

EPMDJ-10732

### DIFF
--- a/admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthMapperImpl.kt
+++ b/admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthMapperImpl.kt
@@ -55,17 +55,12 @@ class OAuthMapperImpl(oauth2Config: OAuth2Config) : OAuthMapper {
     }
 
     private fun mapRole(roleNames: List<String>?): Role {
+        val lowerRoleNames = roleNames?.map { it.lowercase() } ?: emptyList()
         val userRoles = roleMapping.user.split(",").map { it.trim().lowercase() }
         val adminRoles = roleMapping.admin.split(",").map { it.trim().lowercase() }
 
-        roleNames?.forEach { roleName ->
-            val lowerRoleName = roleName.lowercase()
-            if (lowerRoleName in userRoles) {
-                return Role.USER
-            } else if (lowerRoleName in adminRoles) {
-                return Role.ADMIN
-            }
-        }
+        if (adminRoles.any { it in lowerRoleNames }) return Role.ADMIN
+        if (userRoles.any { it in lowerRoleNames }) return Role.USER
         return Role.UNDEFINED
     }
 }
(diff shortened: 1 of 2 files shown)


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthMapperImpl.kt
```diff
@@ -55,10 +55,15 @@ class OAuthMapperImpl(oauth2Config: OAuth2Config) : OAuthMapper {
     }
 
     private fun mapRole(roleNames: List<String>?): Role {
-        roleNames?.forEach {
-            when (it.lowercase()) {
-                roleMapping.user.lowercase() -> return Role.USER
-                roleMapping.admin.lowercase() -> return Role.ADMIN
+        val userRoles = roleMapping.user.split(",").map { it.trim().lowercase() }
```
RomanDavlyatshin, line 59: I think `application.conf` syntax supports lists in the form of 
```
value = [ "foo", "bar", "baz"]
```
Should we just use those instead of comma-separated string?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthMapperImpl.kt
```diff
@@ -55,10 +55,15 @@ class OAuthMapperImpl(oauth2Config: OAuth2Config) : OAuthMapper {
...
+        val adminRoles = roleMapping.admin.split(",").map { it.trim().lowercase() }
+
+        roleNames?.forEach { roleName ->
+            val lowerRoleName = roleName.lowercase()
+            if (lowerRoleName in userRoles) {
+                return Role.USER
+            } else if (lowerRoleName in adminRoles) {
+                return Role.ADMIN
```
RomanDavlyatshin: we should search in adminRoles first for admin role mapping to take precedence

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/OAuthMapperTest.kt
```diff
@@ -223,4 +223,38 @@ class OAuthMapperTest {
...
+        }))
+        val accessToken = JWT.create()
+            .withSubject("some-username")
+            .withClaim("roles", listOf("Ops"))
+            .sign(testAlgorithm)
+
+        val userEntity = oauthMapper.mapAccessTokenPayloadToUserEntity(accessToken)
+
+        assertEquals(Role.ADMIN.name, userEntity.role)
+    }
```
RomanDavlyatshin, line 259: Add test for admin role precedence

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/OAuthMapperImpl.kt
RomanDavlyatshin: We'd like to pass these values through env variables, hence the issue

PR #369, opened 2023-12-21
Feature/cache api keys epmdj 10750

bodyangug:


admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
```diff
@@ -34,6 +36,10 @@ class ApiKeyAuthenticationProvider internal constructor(
     internal val authenticationFunction = config.authenticationFunction
     internal val challengeFunction = config.challengeFunction
     internal val authScheme = config.authScheme
+    internal val cache = Caffeine.newBuilder()
+        .maximumSize(config.maximumCacheSize)
+        .expireAfterWrite(Duration.ofDays(30))
```
iryabov: Why 30 days, why not get it from the config?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
```diff
@@ -56,6 +62,11 @@ class ApiKeyAuthenticationProvider internal constructor(
          */
         var headerName: String = API_KEY_HEADER
 
+        /**
+         * The size that will be used for the cache.
```
iryabov: This doesn't explain how to use it. Not by whom it is used, but how or for what

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
```diff
@@ -56,6 +62,11 @@ class ApiKeyAuthenticationProvider internal constructor(
...
+         */
+        var maximumCacheSize: Long = 1000
```
iryabov: Why 1000?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
```diff
@@ -86,7 +97,9 @@ fun Authentication.Configuration.apiKey(
 
     provider.pipeline.intercept(AuthenticationPipeline.RequestAuthentication) { context ->
         val apiKey = context.call.request.header(provider.headerName)
-        val principal = apiKey?.let { provider.authenticationFunction(context.call, it) }
+
+        val principal = runCatching { provider.cache.getIfPresent(apiKey) }.getOrNull()
```
iryabov: - `runCatching`: It is better to avoid catching situations that can be avoided by a simple check. TBD
- Need to think about how to disable the cache

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
```diff
@@ -100,6 +113,8 @@ fun Authentication.Configuration.apiKey(
         } else if (principal == null) {
             challenge(AuthenticationFailedCause.InvalidCredentials)
         } else {
+            //TODO put each times?
```
iryabov: If cache is enabled, then yes, always when apiKey != null and principal != null

iryabov (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
iryabov: 1h by default like jwt expiration

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
bodyangug: refactored

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
bodyangug: done

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
bodyangug: rewrited

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
bodyangug: moved that value to ApiKeysConfig

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyAuthenticationProvider.kt
bodyangug: as a workaround to disable caching we can set up cacheSize to 0.
Info was taken from docs:
"When size is zero, elements will be evicted immediately after being loaded into the cache. This can be useful in testing, or to disable caching temporarily without a code change. As eviction is scheduled on the configured executor, tests may instead prefer to configure the cache to execute tasks directly on the same thread."

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyConfig.kt
```diff
@@ -27,4 +27,19 @@ class ApiKeyConfig(private val config: ApplicationConfig) {
...
+     */
+    val maximumCacheSize: Long
+        get() = config.propertyOrNull("maximumCacheSize")?.getString()?.toLong() ?: 1000
+
+    /**
+     * The duration in hours after which the stored data in the cache becomes invalid or expired.
+     * By default, 1 hour.
+     */
+    val ttlCacheInHours: Long
+        get() = config.propertyOrNull("ttlCacheInHours")?.getString()?.toLong() ?: 1
```
iryabov: It seems to me that it is better to measure the cache lifetime in minutes rather than in hours, it will be more granular. JWT token expiration is in minutes too.

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyModule.kt
```diff
@@ -52,17 +56,28 @@ val apiKeyServicesDIModule = DI.Module("apiKeyServices") {
...
  * A Ktor Authentication configuration for API key authentication.
  */
 fun Authentication.Configuration.configureApiKeyAuthentication(di: DI) {
     val apiKeyService by di.instance<ApiKeyService>()
+    val cacheService by di.instance<CacheService>()
 
     apiKey("api-key") {
         validate { apiKey ->
-            apiKeyService.signInThroughApiKey(apiKey).toPrincipal()
+            cacheService.getFromCacheOrPutIfAbsent(apiKey){
```
iryabov: Reformat code style `){`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/CacheService.kt
```diff
@@ -0,0 +1,21 @@
+package com.epam.drill.admin.auth.service
+
+import io.ktor.auth.*
+
+/**
+ *  A service that manages the storage and retrieval of apiKeys.
+ */
+interface CacheService {
```
iryabov: If this service is only assotiated with ApiKey why not rename it into `ApiKeyCacheService` ?

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/CacheServiceImpl.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.epam.drill.admin.auth.service.CacheService
+import com.github.benmanes.caffeine.cache.Cache
+import io.ktor.auth.*
+
+/**
+ * Implementation of [CacheService], providing caching functionalities.
+ *
+ * @property cache An instance of Caffeine`s [Cache] that stores key-value pairs, where keys are of type [String] and values are [Principal] objects.
+ */
+class CacheServiceImpl(
```
iryabov: In theory, we can have more than one implementation of a caching service. And it would be more logical to use their cache provider in the name. For this case it would be `CoffeineCacheService`

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
```diff
@@ -0,0 +1,48 @@
...
+            Caffeine.newBuilder()
+                .maximumSize(5)
+                .expireAfterWrite(Duration.ofMinutes(1))
+                .build()
+        )
+
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { User(1, "test1", Role.USER) }
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-2") { User(2, "test2", Role.USER) }
+
+        Assert.assertEquals(2, cacheService.cache.estimatedSize())
```
iryabov: Instead of `org.junit.Assert.assertEquals` it is better to use `kotlin.test.*` asserts since we are already using `kotlin.test.Test`

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
```diff
@@ -0,0 +1,48 @@
...
+                .expireAfterWrite(Duration.ofMinutes(1))
+                .build()
+        )
+
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { user }
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { user }
+
+        Assert.assertEquals(1, cacheService.cache.estimatedSize())
+    }
+
```
iryabov: Add test cases with nullable principle

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/CacheServiceImpl.kt
```diff
@@ -0,0 +1,26 @@
...
+    val cache: Cache<String, Principal>
```
iryabov: Will be better make `cache` private. For test cases, we will be able to obtain a cache provider instance in another way

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
```diff
@@ -0,0 +1,48 @@
...
+            Caffeine.newBuilder()
+                .maximumSize(5)
+                .expireAfterWrite(Duration.ofMinutes(1))
+                .build()
+        )
+
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { user }
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { user }
+
+        Assert.assertEquals(1, cacheService.cache.estimatedSize())
```
iryabov: Instead of getting the cache instance from the `cacheService.cache` service, we can do this by moving the caffeine instance to a variable.

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
```diff
@@ -0,0 +1,48 @@
...
+        )
+
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { User(1, "test1", Role.USER) }
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-2") { User(2, "test2", Role.USER) }
+
+        Assert.assertEquals(2, cacheService.cache.estimatedSize())
+    }
+
+    @Test
+    fun `given only one apiKey two times, then cache size should be one`(): Unit = runBlocking {
```
iryabov: I'm not as good at writing tests as @RomanDavlyatshin , but after going through a lot of reviews from him, I can give some advice on how to improve this test.

The purpose of testing is not to control the cache size, but the fact that we do not access the API key service if the result is already in the cache. 
Regarding this purpose will be better rename test case like 
`if api key is already in cache, getFromCacheOrPutIfAbsent function must return user from cache`.
And we will have to rework the test code so that the `getFromCacheOrPutIfAbsent` function is called only once.
Like this:
```
var testApiKey = "test-api-key"
val testUser = ...
val cache = Coffeine. ... .build()
  .also { put(testApiKey, testUser) }

val user = cacheService.getFromCacheOrPutIfAbsent(testApiKey) { assertFail("Must not be called") }

assertEquals(testUser, user)
```

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.time.Duration
+import kotlin.test.Test
+
+/**
+ * Tests for [CacheServiceImpl] logic
+ */
+class CacheServiceImplTest {
+
+    @Test
+    fun `given two distinct apiKeys, then cache size should be two`(): Unit = runBlocking {
```
iryabov: It would be better to rework this test taking into account the comment for the test below

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyConfig.kt
bodyangug: changed

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyModule.kt
bodyangug: done

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/CacheService.kt
bodyangug: renamed

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/CacheServiceImpl.kt
bodyangug: renamed

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
bodyangug: changed

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
bodyangug: added

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/CacheServiceImpl.kt
bodyangug: done

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
bodyangug: done

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
bodyangug: Refactored. The changes are open for discussion and review.

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CacheServiceImplTest.kt
bodyangug: Refactored. The changes are open for discussion and review.

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -3,46 +3,55 @@ package com.epam.drill.admin.auth
...
+        Mockito.verify(caffeineMock, times(1)).put(key1, user1)
+        assertNotNull(principal)
     }
 
+    @Test
+    fun `if api key is already in cache, getFromCacheOrPutIfAbsent function must return user from cache`(): Unit =
+        runBlocking {
+            val user = User(1, "test1", Role.USER)
+            val key = "custom-key-1"
+            Mockito.`when`(caffeineMock.getIfPresent(key)).thenReturn(user)
```
iryabov: Better to use Mockito Kotlin function `whenever` instead of `when`

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -3,46 +3,55 @@ package com.epam.drill.admin.auth
...
+
+            val principal = cacheService.getFromCacheOrPutIfAbsent(key) {
+                fail(message = "Must not be called")
+            }
+
+            assertNotNull(principal)
```
iryabov: Why not check that the user being returned is equal to the user in the cache?

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -3,46 +3,55 @@ package com.epam.drill.admin.auth
...
-
-        assertEquals(1, cacheService.cache.estimatedSize())
+    fun `given apiKey, then should be called put method`(): Unit = runBlocking {
+        val key1 = "custom-key-1"
+        val user1 = User(1, "test1", Role.USER)
+
+        val principal = cacheService.getFromCacheOrPutIfAbsent(key1) { user1 }
+
+        Mockito.verify(caffeineMock, times(1)).put(key1, user1)
+        assertNotNull(principal)
```
iryabov: Why not check that the user being returned is equal to the user in the function `{ user1 }`?

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -3,46 +3,55 @@ package com.epam.drill.admin.auth
...
-                .expireAfterWrite(Duration.ofMinutes(1))
-                .build()
-        )
-
-        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { user }
-        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { user }
-
-        assertEquals(1, cacheService.cache.estimatedSize())
+    fun `given apiKey, then should be called put method`(): Unit = runBlocking {
+        val key1 = "custom-key-1"
```
iryabov: Better to rename `key` -> `key`. `key1` is not sense.

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -3,46 +3,55 @@ package com.epam.drill.admin.auth
...
+        val user1 = User(1, "test1", Role.USER)
```
iryabov: Better to rename user1 -> user. Any way `user2` is not exists

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -3,46 +3,55 @@ package com.epam.drill.admin.auth
...
+
+        val principal = cacheService.getFromCacheOrPutIfAbsent(key1) { user1 }
+
+        Mockito.verify(caffeineMock, times(1)).put(key1, user1)
```
iryabov: `Mockito.` can be move to static import, so the code will be clearer.

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -3,46 +3,55 @@ package com.epam.drill.admin.auth
...
-                .maximumSize(5)
-                .expireAfterWrite(Duration.ofMinutes(1))
-                .build()
-        )
-
-        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { user }
-        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { user }
-
-        assertEquals(1, cacheService.cache.estimatedSize())
+    fun `given apiKey, then should be called put method`(): Unit = runBlocking {
```
iryabov: Better to rename `if api key is not in cache, getFromCacheOrPutIfAbsent must return user from callback function and put it in cache`

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -54,4 +54,10 @@ class CoffeineCacheServiceTest {
             assertNotNull(principal)
         }
 
+    @Test
+    fun `given principle with value null, then cache must not populate`(): Unit = runBlocking {
```
iryabov: Better to rename `given null principle in callback function, getFromCacheOrPutIfAbsent must return null and cache must not populate`

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
```diff
@@ -54,4 +54,10 @@ class CoffeineCacheServiceTest {
...
+        cacheService.getFromCacheOrPutIfAbsent("custom-key-1") { null }
```
iryabov: Add assert that the return value is null.

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: done

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: added

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: added

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: renamed

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: renamed

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: removed

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: renamed

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: renamed

admin-auth/src/test/kotlin/com/epam/drill/admin/auth/CoffeineCacheServiceTest.kt
bodyangug: added

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyConfig.kt
```diff
@@ -27,4 +27,19 @@ class ApiKeyConfig(private val config: ApplicationConfig) {
...
+
+    /**
+     * The maximum number of items or data size that the cache can hold before it starts replacing or removing existing entries.
+     * By default, 1000.
+     */
+    val maximumCacheSize: Long
+        get() = config.propertyOrNull("maximumCacheSize")?.getString()?.toLong() ?: 1000
+
+    /**
+     * The duration in hours after which the stored data in the cache becomes invalid or expired.
```
RomanDavlyatshin: `The duration in hours` -> `Duration in minutes`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/ApiKeyCacheService.kt
```diff
@@ -0,0 +1,36 @@
...
+/**
+ *  A service that manages the storage and retrieval of apiKeys.
+ */
+interface ApiKeyCacheService {
+    /**
+     * Retrieves a [Principal] object associated with the given [apiKey] from the cache.
+     * If the key is not present in the cache, it invokes the [ifMissCallback] function to retrieve the value,
+     * suspends during its execution, and then stores the result in the cache before returning it.
+     *
+     * @param apiKey The key used to look up the value in the cache.
```
RomanDavlyatshin: Poor grammar; hard to understand the meaning behind the sentence

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/CoffeineCacheService.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.epam.drill.admin.auth.service.ApiKeyCacheService
+import com.github.benmanes.caffeine.cache.Cache
+import io.ktor.auth.*
+
+/**
+ * Implementation of [ApiKeyCacheService], providing caching functionalities.
+ *
+ * @property cache An instance of Caffeine`s [Cache] that stores key-value pairs, where keys are of type [String] and values are [Principal] objects.
+ */
+class CoffeineCacheService(
```
RomanDavlyatshin: `Coffeine` -> `Caffeine`
Keep naming consistent with the library (make sure to check with `CTRL+SHIFT+F` to replace all cases of incorrect spelling)

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/CoffeineCacheService.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val cache: Cache<String, Principal>
+) : ApiKeyCacheService {
+    override suspend fun getFromCacheOrPutIfAbsent(
+        apiKey: String,
+        ifMissCallback: suspend (String) -> Principal?
+    ): Principal? {
+        var principal = cache.getIfPresent(apiKey)
+        if (principal == null) {
+            principal = ifMissCallback.invoke(apiKey)
+            principal?.let { cache.put(apiKey, principal) }
```
RomanDavlyatshin: TBD concurrency

RomanDavlyatshin (changes requested):

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/ApiKeyCacheService.kt
RomanDavlyatshin: `The key used to` -> `The key is used to`

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/CoffeineCacheService.kt
RomanDavlyatshin: it is backed by caffein's Cache which in turn is backed by COncurrentLinkedHashMap

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/config/ApiKeyConfig.kt
bodyangug: renamed

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/ApiKeyCacheService.kt
bodyangug: renamed

admin-auth/src/main/kotlin/com/epam/drill/admin/auth/service/impl/CoffeineCacheService.kt
bodyangug: renamed

PR #376, opened 2024-02-01
Feature/review configuration classes try to remove custom url class epmdj 10659

epamrsa:


admin-core/src/main/kotlin/com/epam/drill/admin/endpoints/instance/AgentInstanceEndpoints.kt
```diff
@@ -77,7 +77,7 @@ class AgentInstanceEndpoints(override val di: DI) : DIAware {
 
     private fun Route.registerAgentInstanceRoute() {
         put<Agents> {
-            val agentConfig = call.decompressAndReceive<AgentConfig>()
+            val agentConfig = call.decompressAndReceive<AgentMetadata>()
```
RomanDavlyatshin: TODO rename agentConfig to `agentMetadata`

PR #377, opened 2024-02-01
Refactor/ rewrite AgentInstanceEndpoints routes to new style

bodyangug:


admin-core/src/main/kotlin/com/epam/drill/admin/endpoints/instance/AgentInstanceEndpoints.kt
```diff
@@ -61,92 +63,80 @@ object Agents {
...
-    private fun Route.completeSendingClassMetadataRoute() {
-        post<Agents.ClassMetadataComplete> { params ->
-            handleAgentRequest(params.agentId, params.buildVersion) { agentInfo ->
-                processPluginData(agentInfo, Initialized())
-                call.addDrillInternalHeader()
+            post<Agents.ClassMetadata> { params ->
+                handleAgentRequest(agentManager, params.agentId, params.buildVersion) { agentInfo ->
+                    val data = call.decompressAndReceive<ClassMetadata>()
+                    processPluginData(pd, agentInfo, data.toInitDataPart())
+                    call.addDrillInternalHeader()
```
iryabov: It is better to place the `drill-internal` header inside the `handleAgentRequest` function, and send it to the agent in all cases, not only in case of success.

bodyangug: done

admin-core/src/main/kotlin/com/epam/drill/admin/endpoints/instance/AgentInstanceEndpoints.kt
```diff
@@ -61,90 +63,76 @@ object Agents {
...
+                else -> {
+                    logger.error { "The active build version of agent $agentId is ${agentInfo.build.version}, but the request has $buildVersion build version!" }
+                    call.respond(HttpStatusCode.Conflict)
                 }
-            } ?: call.respond(HttpStatusCode.NotFound).also {
-            logger.error { "Agent $agentId is not attached!" }
-        }
+            }
+        } ?: call.respond(HttpStatusCode.NotFound).also {
+        logger.error { "Agent $agentId is not attached!" }
```
iryabov, line 126: It seems this branch won't return header `drill-internal`

bodyangug: fixed

iryabov (approved):

PR #382, opened 2024-03-14
Feature/varbit exposed epmdj 10811

iryabov:


admin-app/src/main/kotlin/com/epam/drill/admin/DrillAdminApplication.kt
```diff
@@ -16,111 +16,103 @@
...
+            //Data
+            authenticate("api-key") {
+                withRole(Role.USER, Role.ADMIN) {
+                    intercept(ApplicationCallPipeline.Call) {
+                        call.response.header("drill-internal", "true")
+                        proceed()
+                    }
+                    putAgentConfig()
+                    postCoverage()
+                    postCLassMetadata()
```
RomanDavlyatshin: `postCLass` typo

admin-app/src/main/kotlin/com/epam/drill/admin/config/DatabaseConfig.kt
```diff
@@ -0,0 +1,78 @@
...
+import io.ktor.config.*
+import org.kodein.di.DI
+import org.kodein.di.bind
+import org.kodein.di.instance
+import org.kodein.di.singleton
+import javax.sql.DataSource
+
+class DatabaseConfig(private val config: ApplicationConfig) {
+    val host: String
+        get() = config.propertyOrNull("host")?.getString() ?: "localhost"
```
RomanDavlyatshin, line 30: reason for using getter instead of initializing assignment

RomanDavlyatshin, line 30: As far as I remember, we intentionally decided not to create defaults "everywhere", and instead keep default values to a single configuration file - `application.conf`

admin-app/src/main/kotlin/com/epam/drill/admin/config/DatabaseConfig.kt
```diff
@@ -0,0 +1,78 @@
...
+        val maxPoolSize = databaseConfig.maxPoolSize
+
+        HikariConfig().apply {
+            this.driverClassName = "org.postgresql.Driver"
+            this.jdbcUrl = "jdbc:postgresql://$host:$port/$dbName"
+            this.username = userName
+            this.password = password
+            this.maximumPoolSize = maxPoolSize
+            this.isAutoCommit = true
+            this.transactionIsolation = "TRANSACTION_REPEATABLE_READ"
```
RomanDavlyatshin, line 69: reason for setting REPEATABLE_READ

admin-app/src/main/kotlin/com/epam/drill/admin/config/DatabaseConfig.kt
```diff
@@ -0,0 +1,78 @@
...
+        val password = databaseConfig.password
+        val maxPoolSize = databaseConfig.maxPoolSize
+
+        HikariConfig().apply {
+            this.driverClassName = "org.postgresql.Driver"
+            this.jdbcUrl = "jdbc:postgresql://$host:$port/$dbName"
+            this.username = userName
+            this.password = password
+            this.maximumPoolSize = maxPoolSize
+            this.isAutoCommit = true
```
RomanDavlyatshin, line 68: we actually don't want `isAutoCommit`
According to chatgpt:


**You**:
postgresql 14
does isAutoCommit = true affect batch/bulk insert operations?

**ChatGPT**:
In PostgreSQL, the concept of "auto-commit" typically refers to whether each statement is automatically committed after execution. When isAutoCommit is set to true, each statement is committed immediately after execution, which means that each batch or bulk insert operation would be committed individually.

If you're performing batch or bulk insert operations and want to ensure that the entire batch is committed as a single transaction, you would typically set isAutoCommit to false and explicitly commit the transaction after all insert operations have been completed. This approach ensures that either all inserts succeed or none are committed, maintaining data consistency.

Here's a basic example in Java using JDBC to illustrate this:

admin-app/src/main/resources/application.conf
```diff
@@ -100,4 +57,7 @@ drill {
             }
         }
     }
+    test2code {
```
RomanDavlyatshin: Reason for keeping empty config section

admin-writer/src/main/kotlin/com/epam/drill/admin/writer/rawdata/config/BitSetColumnType.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.postgresql.util.PGobject
+import java.util.*
+
+class BitSetColumnType : ColumnType() {
+    override fun sqlType(): String = "VARBIT"
+
+    override fun valueFromDB(value: Any): Any =
+        when (value) {
+            is String -> value.toBitSet()
+            is PGobject -> (value.value ?: "").toBitSet()
```
RomanDavlyatshin, line 28: case for `PGobject` vs `String`

RomanDavlyatshin (changes requested):

admin-app/src/main/kotlin/com/epam/drill/admin/config/DatabaseConfig.kt
RomanDavlyatshin: We had overcomplications with `application.conf` due to Auth impl - it is no longer a single source of truth

RomanDavlyatshin: required for hot-reloading of config during development (ktor feature)

admin-writer/src/main/kotlin/com/epam/drill/admin/writer/rawdata/config/BitSetColumnType.kt
RomanDavlyatshin: both lines are covered, we don't know case clearly yet

RomanDavlyatshin (approved):

PR #388, opened 2024-05-28
fix: use `decompressAndReceive` function in routes to explicitly deco…

iryabov:
…mpress the request body

### DIFF
--- a/admin-writer/src/main/kotlin/com/epam/drill/admin/writer/rawdata/route/RawDataWriterRoutes.kt
+++ b/admin-writer/src/main/kotlin/com/epam/drill/admin/writer/rawdata/route/RawDataWriterRoutes.kt
@@ -16,22 +16,29 @@
 package com.epam.drill.admin.writer.rawdata.route
 
 import com.epam.drill.admin.writer.rawdata.service.RawDataWriter
-import io.ktor.server.application.*
 import io.ktor.client.*
 import io.ktor.client.engine.apache.*
+import io.ktor.client.plugins.contentnegotiation.*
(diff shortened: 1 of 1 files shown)


admin-writer/src/main/kotlin/com/epam/drill/admin/writer/rawdata/route/RawDataWriterRoutes.kt
```diff
@@ -126,4 +133,34 @@ internal suspend fun sendPostRequest(url: String, data: Any) {
...
+internal suspend inline fun <reified T : Any> ApplicationCall.decompressAndReceive(): T {
+    val body: ByteArray = when (request.headers[HttpHeaders.ContentEncoding]) {
+        "gzip" -> decompressGZip(receiveStream())
+        else -> receive<ByteArray>()
+    }
+    return when (request.headers[HttpHeaders.ContentType]) {
+        ContentType.Application.ProtoBuf.toString() -> ProtoBuf.decodeFromByteArray(T::class.serializer(), body)
+        ContentType.Application.Json.toString() -> json.decodeFromString(T::class.serializer(), String(body))
+        else -> throw UnsupportedMediaTypeException(
+            ContentType.parse(request.headers[HttpHeaders.ContentType] ?: "application/octet-stream")
```
RomanDavlyatshin: Defaulting to `?: "application/octet-stream"` is misleading. I suggest replacing it with `?: "Content-Type header is missing"`

RomanDavlyatshin (changes requested):

PR #402, opened 2024-11-25
feat: add /tests-to-skip route and a stub implementation

iryabov:
EPMDJ-10877


admin-metrics/src/main/kotlin/com/epam/drill/admin/metrics/route/response/TestResponse.kt
RomanDavlyatshin, line 1: I suggest we keep it similar to `admin-writer` implementation  - we can use entity from `admin-writer/src/main/kotlin/com/epam/drill/admin/writer/rawdata/entity/TestMetadata.kt` and introduce `/view/TestToSkipView`

RomanDavlyatshin (changes requested):

admin-writer/src/main/resources/raw_data/db/migration/R__Metrics_Functions.sql
```diff
@@ -1413,3 +1413,136 @@ BEGIN
...
+    __test_runner_id VARCHAR,
+    __test_path VARCHAR,
+    __test_name VARCHAR
+) AS $$
+BEGIN
+    RETURN QUERY
+    WITH
+    Risks AS (
+        SELECT
+			risks.__signature AS signature,
```
RomanDavlyatshin: formatting is off

admin-writer/src/main/resources/raw_data/db/migration/R__Metrics_Functions.sql
```diff
@@ -1413,3 +1413,136 @@ BEGIN
...
+		    risks.__risk_type AS risk_type,
+            risks.__name AS risk_name,
+            risks.__body_checksum AS risk_body_checksum
+        FROM raw_data.get_risks(input_target_build_id, input_baseline_build_id) risks
+    ),
+    TargetMethods AS (
```
RomanDavlyatshin: Can we use `raw_data.get_methods` fn here?

admin-writer/src/main/resources/raw_data/db/migration/R__Metrics_Functions.sql
```diff
@@ -1413,3 +1413,136 @@ BEGIN
...
+                    OR (r.classname_pattern IS NOT NULL AND methods.classname ~ r.classname_pattern)
+                    OR (r.annotations_pattern IS NOT NULL AND methods.annotations ~ r.annotations_pattern)
+                    OR (r.class_annotations_pattern IS NOT NULL AND methods.class_annotations ~ r.class_annotations_pattern))
+                    AND r.group_id = input_group_id
+                    AND r.app_id = input_target_app_id
+            )
+  	),
+  	MethodCoverage AS (
+        SELECT
+            methods.signature as signature,
```
RomanDavlyatshin: no need to explicitly add `as` for each column if name is not changed

admin-writer/src/main/resources/raw_data/db/migration/R__Metrics_Functions.sql
```diff
@@ -1413,3 +1413,136 @@ BEGIN
...
+        JOIN raw_data.instances instances ON instances.id = coverage.instance_id
+        JOIN raw_data.methods methods ON methods.classname = coverage.classname AND methods.build_id = instances.build_id
+        WHERE BIT_COUNT(SUBSTRING(coverage.probes FROM methods.probe_start_pos + 1 FOR methods.probes_count)) > 0
+  	),
+	Coverage AS (
+		SELECT
+		    def.id AS test_definition_id,
+			launches.id AS test_id,
+			builds.id AS build_id,
+			(target.body_checksum != coverage.body_checksum) AS has_risk
```
RomanDavlyatshin: aliases made this condition very confusing

RomanDavlyatshin: `has_risk` is never checked to be "false" -> all the records containing "false" values are never used -> can't we just filter them right away with JOIN condition?

iryabov: The column `hasRisk` is used in `TestsByRisks` CTE

admin-metrics/src/main/kotlin/com/epam/drill/admin/metrics/route/response/TestResponse.kt
iryabov: How about moving common models as `TestMetadata` to the new `admin-common` module?

admin-writer/src/main/resources/raw_data/db/migration/R__Metrics_Functions.sql
```diff
@@ -1413,3 +1413,824 @@ BEGIN
...
+			methods.probes_count
+	 	FROM raw_data.view_methods_with_rules methods
+	 	WHERE methods.build_id = input_target_build_id
+  	),
+	Coverage AS (
+		SELECT
+		    coverage.test_definition_id,
+			coverage.test_id,
+			coverage.build_id,
+			(target.body_checksum != coverage.body_checksum) AS has_risk
```
RomanDavlyatshin: `has_risk` -> `is_modified`

admin-writer/src/main/resources/raw_data/db/migration/V6__Methods_Coverage_Mat_View.sql
RomanDavlyatshin, line 1: Change V6__ to R__Mat_Views to avoid spamming V__ migrations for mat view changes AND keep mat views separated to avoid unnecessary updates to matviews when only metrics functions change

admin-writer/src/main/resources/raw_data/db/migration/R__Metrics_Functions.sql
RomanDavlyatshin, line 1: R__Metric_Functions
R__Views
R__Matviews

PR #404, opened 2024-12-23
Feature/v2 functions for coverage epmdj 10960

iryabov:


admin-writer/src/main/resources/raw_data/db/migration/R__Metrics_Functions.sql
```diff
@@ -1413,3 +1413,719 @@ BEGIN
...
 $$ LANGUAGE plpgsql;
+
+-----------------------------------------------------------------
+
+-----------------------------------------------------------------
+CREATE OR REPLACE VIEW raw_data.view_methods_coverage AS
+    SELECT
+        builds.group_id,
+        builds.app_id,
+        methods.signature as signature,
```
RomanDavlyatshin: lets remove `as ...` - we don't change column names here, no reason to add these

PR #409, opened 2025-01-20
Feature/data retention epmdj 10969

iryabov:


admin-app/src/main/resources/application.conf
```diff
@@ -68,11 +68,12 @@ drill {
...
-        scheduler {
-            refreshViewsIntervalInMinutes = 30
-            refreshViewsIntervalInMinutes = ${?DRILL_METRICS_SCHEDULER_REFRESH_VIEWS_INTERVAL_IN_MINUTES}
-        }
     }
     testRecommendations {
     }
+    scheduler {
+        refreshViewsIntervalInMinutes = 30
+        refreshViewsIntervalInMinutes = ${?DRILL_METRICS_SCHEDULER_REFRESH_VIEWS_INTERVAL_IN_MINUTES} //todo DRILL_METRICS_SCHEDULER_REFRESH_VIEWS_INTERVAL_IN_MINUTES is deprecated, use DRILL_SCHEDULER_REFRESH_VIEWS_INTERVAL_IN_MINUTES
```
iryabov: I moved the `scheduler` property according to the current location of the scheduler code. But I decided to keep the old env variable for backward compatibility since we have clients

RomanDavlyatshin: we dont have anyone using this yet, so we can change that

PR #412, opened 2025-02-04
Feature/save test finishing time epmdj 10980

iryabov:


admin-writer/src/main/resources/raw_data/db/migration/V12__Test_Duration.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE raw_data.test_launches
+ADD COLUMN IF NOT EXISTS duration BIGINT DEFAULT NULL;
```
RomanDavlyatshin: BIGINT is an overkill, integer is sufficient
https://www.postgresql.org/docs/current/datatype-numeric.html

admin-writer/src/main/kotlin/com/epam/drill/admin/writer/rawdata/service/impl/RawDataServiceImpl.kt
```diff
@@ -192,7 +191,7 @@ class RawDataServiceImpl(
             id = sessionPayload.id,
             groupId = sessionPayload.groupId,
             testTaskId = sessionPayload.testTaskId,
-            startedAt = sessionPayload.startedAt.toLocalDateTime(TimeZone.UTC).toJavaLocalDateTime()
+            startedAt = sessionPayload.startedAt.toLocalDateTime(TimeZone.currentSystemDefault()).toJavaLocalDateTime()
```
RomanDavlyatshin: What's the motivation for this change?

RomanDavlyatshin (changes requested):

admin-writer/src/main/resources/raw_data/db/migration/V12__Test_Duration.sql
iryabov: I agree that for milliseconds integer will be enough for a test execution time. It's just that the duration of time is usually measured in a long type in Java and a cast to int type will be required

RomanDavlyatshin (approved):

PR #417, opened 2025-02-20
Feature/recommended tests filters epmdj 10984

iryabov:


admin-writer/src/main/resources/raw_data/db/migration/R__3_Functions.sql
```diff
@@ -2141,3 +2159,67 @@ BEGIN
 	FROM Tests;
 END;
 $$ LANGUAGE plpgsql;
+
+-----------------------------------------------------------------
+
+-----------------------------------------------------------------
+CREATE OR REPLACE FUNCTION raw_data.format_duration(milliseconds bigint)
```
RomanDavlyatshin, line 2166: I strongly dislike the fact we're doing this in SQL, but as far as I understand this is used in Metabase queries, is that right?

PR #420, opened 2025-03-03
Feature/optimize aggregate coverage queries epmdj 10994

iryabov:


RomanDavlyatshin (approved):

admin-writer/src/main/resources/raw_data/db/migration/R__1_Views.sql
```diff
@@ -52,31 +51,99 @@ CREATE OR REPLACE VIEW raw_data.view_methods_with_rules AS
...
+              i.build_id
+            FROM
+              raw_data.coverage c
+            JOIN raw_data.test_launches tl ON tl.id = c.test_id
+            JOIN raw_data.instances i ON i.id = c.instance_id
+            WHERE tl.test_session_id = ts.id
+            LIMIT 1
+          ) AS build_id,
+        sum(tl.duration) AS duration,
+        SUM(CASE WHEN tl.result = 'FAILED' THEN 1 ELSE 0 END) AS failed,
```
RomanDavlyatshin, line 105: Why do we want multiple columns with booleans instead of a single field with actual status?

PR #432, opened 2025-04-25
Feature/unique coverage epmdj 11015

iryabov:


admin-app/src/main/kotlin/com/epam/drill/admin/config/SchedulerConfig.kt
```diff
@@ -32,7 +32,7 @@ class SchedulerConfig(private val config: ApplicationConfig) {
     private val refreshViewsIntervalInMinutes: Int =
         config.propertyOrNull("refreshViewsIntervalInMinutes")?.getString()?.toInt() ?: 30
     private val dataRetentionJobCron: String = config.propertyOrNull("dataRetentionJobCron")?.getString() ?: "0 0 1 * * ?"
-    val threadPools: Int = config.propertyOrNull("threadPools")?.getString()?.toInt() ?: 4
+    val threadPools: Int = config.propertyOrNull("threadPools")?.getString()?.toInt() ?: 2
```
RomanDavlyatshin, line 35: whats the motivation for reducing number of threadPools to 2?

admin-writer/src/main/resources/raw_data/db/migration/R__2_Mat_Views.sql
```diff
@@ -185,4 +185,238 @@ SELECT
...
+        SELECT
+            m.build_id,
+            m.signature,
+            m.body_checksum,
+            m.probes_count,
+            m.probes_start,
+            m.method_num,
+            baseline.build_id AS baseline_build_id,
+            baseline.body_checksum AS baseline_body_checksum,
+            (CASE WHEN m.body_checksum = baseline.body_checksum AND m.probes_count = baseline.probes_count THEN 1 ELSE 0 END) AS equal,
```
RomanDavlyatshin, line 364: whats the motivation for checking probes count? Isn't checksum equality check sufficient?

admin-writer/src/main/resources/raw_data/db/migration/R__3_Functions.sql
```diff
@@ -4252,4 +1033,684 @@ BEGIN
...
+			GROUP BY coverage.build_id
+		),
+		ComparableBuildCoverage AS (
+			SELECT
+				coverage.build_id,
+				raw_data.CONCAT_VARBIT(coverage.probes::VARBIT, coverage.probes_start::INT) AS probes
+			FROM ComparableMethodsCoverage coverage
+			GROUP BY coverage.build_id
+		),
+	    AugmentedTargetBuildCoverage AS (
```
RomanDavlyatshin: It's hard to tell what "augmented" means

admin-writer/src/main/resources/raw_data/db/migration/R__2_Mat_Views.sql
iryabov: Yes, it is important to take into account both the method checksum and the probes count, because we collect all the probes of equivalent methods into one long build probe. If the methods have the same checksum, but the number of probes does not match, then the probes of the same build will be of different lengths. This will break the calculation of aggregated coverage for builds.

admin-app/src/main/kotlin/com/epam/drill/admin/config/SchedulerConfig.kt
iryabov: Perhaps - this is a workaround for accounting for materialized view update dependencies. However, it still does not guarantee the correct sequence during parallelization, so it would be correct to do parallelization in the task itself, and set the number of threads based on other considerations. For example, based on considerations of 1 thread per 1 job.
In PR #434 I have already implemented parallelization in the job itself, and there are only 2 jobs left, so you can leave 2 threads here, taking into account that PR

PR #435, opened 2025-07-01
feat: create metrics database schema

iryabov:
feat: migrate all analytical views and functions from `raw_data` to `metrics` database schema

EPMDJ-11023


admin-metrics/src/main/kotlin/com/epam/drill/admin/metrics/service/impl/MetricsServiceImpl.kt
```diff
@@ -470,22 +461,26 @@ class MetricsServiceImpl(
...
+        val buildMethodsViewJob = async { refreshMaterializedView(buildMethodsView) }
+        val testLaunchesViewJob = async { refreshMaterializedView(testLaunchesView) }
+        val testDefinitionsViewJob = async { refreshMaterializedView(testDefinitionsView) }
+        val testSessionsViewJob = async { refreshMaterializedView(testSessionsView) }
+        val methodCoverageViewJob = async { waitFor(buildsViewJob, methodsViewJob, buildMethodsViewJob,
+            testLaunchesViewJob, testDefinitionsViewJob, testSessionsViewJob).run { refreshMaterializedView(methodCoverageView) } }
+        val methodSmartCoverageViewJob = async { waitFor(methodCoverageViewJob).run { refreshMaterializedView(methodSmartCoverageView) } }
+        val testSessionBuildsViewJob = async { waitFor(methodCoverageViewJob).run { refreshMaterializedView(testSessionBuildsView) } }
+
+        waitFor(testSessionBuildsViewJob, methodSmartCoverageViewJob)
```
RomanDavlyatshin, line 475: why wait for these specific jobs, but not the other ones?

admin-metrics/src/main/resources/metrics/db/migration/R__1_Data.sql
```diff
@@ -0,0 +1,208 @@
...
+    CASE WHEN tl.result = 'SMART_SKIPPED' THEN 1 ELSE 0 END AS smart_skipped,
+    CASE WHEN tl.result <> 'FAILED' THEN 1 ELSE 0 END AS success
+FROM raw_data.test_launches tl;
+CREATE UNIQUE INDEX IF NOT EXISTS idx_test_launches_pk ON metrics.test_launches (group_id, test_launch_id);
+
+-----------------------------------------------------------------
+-- Create a materialized view of test definitions
+-- by copying data from the `raw_data.test_definitions` table
+-----------------------------------------------------------------
+DROP MATERIALIZED VIEW IF EXISTS metrics.test_definitions CASCADE;
```
RomanDavlyatshin, line 92: There is not much happening (in terms of transformations/aggregations) in this materialized view - what's our motivation for having it instead of original table?

admin-metrics/src/main/resources/metrics/db/migration/R__3_Functions.sql
```diff
@@ -0,0 +1,796 @@
...
+    input_build_id VARCHAR DEFAULT NULL,
+    input_branches VARCHAR[] DEFAULT NULL
+)
+RETURNS TABLE (
+    group_id VARCHAR,
+    app_id VARCHAR,
+    build_id VARCHAR,
+	target_total_methods BIGINT,
+	target_equal_methods BIGINT,
+	identity_ratio FLOAT
```
RomanDavlyatshin, line 636: Idea: we could use `identity_ratio` threshold to allow filtering out vastly different builds (based on the assumption, that past a certain "similarity" index we can't really apply information from the other build.
That could probably help to improve performance for large apps.
What do you think about it @iryabov ?

admin-metrics/src/main/resources/metrics/db/migration/V1__Metrics_schema_init.sql
```diff
@@ -0,0 +1,57 @@
+CREATE OR REPLACE FUNCTION metrics.format_duration(milliseconds bigint)
```
RomanDavlyatshin, line 1: do we have these functions used somewhere on a critical / highload functions? If yes, I am concerned about performance - we could probably format on the client side (even frontend) w/o issues

admin-metrics/src/main/kotlin/com/epam/drill/admin/metrics/service/impl/MetricsServiceImpl.kt
iryabov: By this point the code has already waited for other jobs, so only the remaining ones are waiting here

admin-metrics/src/main/resources/metrics/db/migration/R__1_Data.sql
iryabov: I considered 2 reasons to do so. 
The first reason is for the data to be consistent. A snapshot of the data at a certain point in time is created so that we do not get data disappearance or, conversely, the appearance of unnecessary data.
The second reason is to create a reliable time window for metrics. We simply will not copy old data in these table snapshots, and we can safely use these tables in any views and functions without worrying about the time window anymore.

admin-metrics/src/main/resources/metrics/db/migration/V1__Metrics_schema_init.sql
iryabov: These functions are used only in Metabase.

PR #437, opened 2025-07-07
feat: add test_session_id to raw_data.coverage

RomanDavlyatshin:
TODO: fix tests


iryabov:
I would add another migration script to update the test_session_id field for all coverage records using the test_id field and the reference from test runs to test sessions.

PR #445, opened 2025-11-06
Feature/metrics etl processing

iryabov:


admin-etl/src/main/kotlin/com/epam/drill/admin/etl/EtlLoadingResult.kt
```diff
@@ -0,0 +1,71 @@
...
+    override val duration: Long? = null,
+    override val errorMessage: String? = null
+): BatchResult(success, duration, errorMessage), Comparable<EtlLoadingResult> {
+    companion object {
+        val EMPTY = EtlLoadingResult(
+            success = true
+        )
+    }
+
+    operator fun plus(other: EtlLoadingResult): EtlLoadingResult {
```
RomanDavlyatshin, line 30: 1. What are the use cases for summing process results?
2. When we would want to combine successful and failed loading results?

admin-etl/src/main/kotlin/com/epam/drill/admin/etl/EtlStatus.kt
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.epam.drill.admin.etl
+
+enum class EtlStatus {
+    SUCCESS,
+    FAILURE,
+    NEVER_RUN
```
RomanDavlyatshin: `NEVER_RUN` -> `NEVER_RAN`, or maybe even clearer - `NEVER_EXECUTED`

admin-etl/src/test/kotlin/com/epam/drill/admin/etl/ETLSimpleTest.kt
```diff
@@ -0,0 +1,221 @@
...
+        val orchestrator = simpleOrchestrator
+
+        // Add initial data
+        addNewRecords(5)
+        // First run ETL — should process all initial data
+        val result1 = orchestrator.run()
+        assertTrue(result1.first().success)
+        assertEquals(5, result1.first().rowsProcessed)
+
+        Thread.sleep(10)
```
RomanDavlyatshin, line 276: ChatGPT suggest that this test can be flaky in an interesting way:
1. `Thread.sleep()` has millisecond precision
2. `Instant.now()` has nanosecond precision, but operates based on a system clock
3. Some implementations of JVM do not guarantee that subsequent calls of `Instant.now()` would actually return different timestamps, as system clock updates can be less frequent (e.g. happen every 15ms) - so consecutive calls  _can_ return equal timestamps.

That sounds like a big problem for ETL in general, since sort order purely based of `Instant.now()` won't be stable

Further it suggests:
In ETL pipelines where **time-based ordering is critical**, implementations typically handle it in a few standard ways:

| Strategy                                          | How it works                                                                         | Pros                                                        | Cons                                                             |
| ------------------------------------------------- | ------------------------------------------------------------------------------------ | ----------------------------------------------------------- | ---------------------------------------------------------------- |
| **Use a strict sort key (timestamp + unique ID)** | Instead of just `createdAt`, use `(timestamp, primaryKey)` for ordering              | Guarantees deterministic order even if timestamps are equal | Slightly more complex queries                                    |
| **Database ordering in SQL**                      | `ORDER BY created_at, id` in the extractor query                                     | Leverages DB sorting; no reliance on `Flow` order           | DB may still return batches; need to preserve order in streaming |
| **Monotonically increasing synthetic timestamp**  | ETL adds a sequence or monotonic counter to records as they are extracted            | Ensures uniqueness in ordering                              | Adds extra field to schema                                       |
| **Buffered batching**                             | Collect a batch of rows, sort in memory before emitting to downstream loaders        | Keeps streaming but enforces stable order                   | Higher memory usage per batch                                    |
| **Idempotent loading**                            | If order is not strictly guaranteed, ensure loaders can handle duplicates/reordering | Simplifies ETL logic                                        | Might not suit analytics pipelines needing exact order           |

**Key points:**

* **Never rely solely on `Instant.now()`** for ordering in ETL; system clock granularity and batch fetching can break strict ordering.
* Most production ETLs use a combination of **timestamp + unique ID** or **DB-side ordering** to ensure deterministic results.
* Sorting **inside the ETL** after extraction is common when streaming rows in batches.

Lets discuss it

RomanDavlyatshin (commented):
I have looked through the `admin-etl` module including tests and left some comments to talk about.

I'll look at `admin-metrics` in the next review

admin-etl/src/test/kotlin/com/epam/drill/admin/etl/ETLSimpleTest.kt
RomanDavlyatshin: Not an issue - we don't care about minute changes in order for our metrics calculation

admin-etl/src/main/kotlin/com/epam/drill/admin/etl/EtlLoadingResult.kt
RomanDavlyatshin: - "summing" is more like "merging"
- we want earliest successful date in order to re-calculate data that was not processed properly

admin-etl/src/main/kotlin/com/epam/drill/admin/etl/impl/UntypedSqlDataExtractor.kt
```diff
@@ -34,13 +34,19 @@ class UntypedSqlDataExtractor(
...
+                        is String -> PGobject().apply {
+                            type = "varbit"
+                            value = bit
+                        }
+                        is PGobject -> bit
                         else -> throw IllegalStateException("Unsupported BIT/VARBIT type: ${bit?.javaClass?.name}")
                     }
                 }
-                "json" -> rs.getObject(i).toString().replace("'", "''")
+//                "json" -> rs.getObject(i).toString().replace("'", "''")
```
RomanDavlyatshin: reason for commenting out quotes escaping

admin-metrics/src/main/resources/metrics/db/etl/build_method_test_definition_coverage_loader.sql
```diff
@@ -23,7 +23,7 @@ SELECT
     :created_at_day,
     :created_at_day,
     :probes
-WHERE :test_session_id IS NOT NULL AND :test_definition_id IS NOT NULL
```
RomanDavlyatshin: motivation for moving that logic to kotlin

admin-app/src/main/resources/openapi.yml
```diff
@@ -914,32 +914,17 @@ paths:
...
+        - name: reset
+          description: Remove all data before start ETL processing
+          in: query
+          required: false
           schema:
-            type: string
+            type: boolean
       responses:
         '200':
           description: Materialized views refresh started
```
RomanDavlyatshin: TODO: cleanup all instances of using "materialized view"

admin-app/src/main/resources/openapi.yml
```diff
@@ -846,6 +918,13 @@ paths:
         - metrics
       security:
         - apiKeyAuth: [ ]
+      parameters:
+        - name: reset
+          description: Remove all data before start ETL processing
```
RomanDavlyatshin: "all data" is not descriptive enough - it could sound like this will simply remove __all the data__, including data collected by agents (raw_data).
Let's think of a better phrasing:
`"If true, clears existing computed data and restarts the ETL process from scratch."`

RomanDavlyatshin (changes requested):

admin-etl/src/main/kotlin/com/epam/drill/admin/etl/impl/UntypedSqlDataExtractor.kt
iryabov: There is no need to deserialize types anymore, since prepared parameters are used.

admin-metrics/src/main/resources/metrics/db/etl/build_method_test_definition_coverage_loader.sql
iryabov: It turned out that simple conditions are processed faster in Kotlin than when using SELECT+WHERE in PostgreSQL, because there is a lot of overhead for transport, serialization, planning, etc.

RomanDavlyatshin (approved):
lgtm 👍

PR #451, opened 2025-12-10
feat: parallelize ETL processing across application groups

iryabov:
EPMDJ-11161


RomanDavlyatshin (approved):
Maybe we should create some unit test for UntypedPreparedSql implementation

PR #453, opened 2026-01-15
feat: run vacuum every N rows per-table

RomanDavlyatshin:


RomanDavlyatshin:
does not solve the issue

PR #462, opened 2026-02-12
Feature/impacted methods in build diff report epmdj 11198

iryabov:


RomanDavlyatshin (approved):

admin-metrics/src/main/kotlin/com/epam/drill/admin/metrics/repository/impl/MetricsRepositoryImpl.kt
```diff
@@ -249,59 +249,110 @@ class MetricsRepositoryImpl : MetricsRepository {
...
-            buildId,
-            baselineBuildId,
-            buildId,
-            baselineBuildId,
-            buildId
-        ).first() as Map<String, String>
+                        COALESCE((SELECT tested_methods FROM TestedChanges WHERE change_type = 'new'), 0) as tested_new_methods,
+                        COALESCE((SELECT tested_methods FROM TestedChanges WHERE change_type = 'modified'), 0) as tested_modified_methods,
+                        (SELECT isolated_probes_coverage_ratio FROM Coverage) as coverage,                                                                        
+                        (SELECT aggregated_probes_coverage_ratio FROM Coverage) as aggregated_coverage,
```
RomanDavlyatshin, line 348: Does aggregation also include information from other branches? AFAIR we wanted to exclude that

admin-metrics/src/main/kotlin/com/epam/drill/admin/metrics/route/MetricRoutes.kt
```diff
@@ -110,7 +110,7 @@ class Metrics {
         val baselineInstanceId: String? = null,
         val baselineCommitSha: String? = null,
         val baselineBuildVersion: String? = null,
-        val coverageThreshold: Double = 1.0, // TODO Float should be enough
+        val coverageThreshold: Double = 0.0,
```
RomanDavlyatshin, line 113: I think we wanted to put that into raw_data.group_settings

PR #466, opened 2026-04-02
feat: add a separate endpoint to persist build info (Git metadata)

iryabov:


admin-app/src/main/resources/openapi.yml
```diff
@@ -98,6 +97,35 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/MessageResponse'
+  /api/data-ingest/builds/info:
+    put:
+      summary: Persist application build Git metadata
+      description: |
+        Saves Git metadata (branch, commitDate, commitMessage, commitAuthor) for an application build.
+        The build is identified by groupId, appId, commitSha, and buildVersion. If a build with the given
+        identity does not exist, a new build is created. If it already exists, only the Git metadata fields
```
RomanDavlyatshin, line 106: whats the usecase for creating new build with that endpoint?
I thought we're going for 2 dedicated endpoints - /build to create build, /build/info to submit git metadata

RomanDavlyatshin: I guess it could be explained with us trying to keep consistent data regardless of order of API calls.
That would've not been an issue though, if we kept build metadata in a separate table
It's fine for now, but lets discuss that

RomanDavlyatshin (approved):

PR #468, opened 2026-04-07
Feature/etl inconsistency window epmdj 11206

iryabov:


admin-app/src/main/resources/application.conf
```diff
@@ -104,5 +104,7 @@ drill {
         transformationBufferSize = ${?DRILL_ETL_TRANSFORMATION_BUFFER_SIZE}
         loggingFrequency = 10
         loggingFrequency = ${?DRILL_ETL_LOGGING_FREQUENCY}
+        consistencyWindow = 0
```
RomanDavlyatshin, line 107: I suggest to set a default value fitting for DRILL_SCHEDULER_ETL_JOB_CRON default value

RomanDavlyatshin (approved):

admin-app/src/main/resources/application.conf
RomanDavlyatshin: irrelevant - consistencyWindow has more to do with maximum test duration rather than etl job frequency

RomanDavlyatshin (approved):

PR #487, opened 2026-07-27
Feature/agent health status epmdj 11240

iryabov:


admin-writer/src/main/kotlin/com/epam/drill/admin/writer/rawdata/repository/impl/InstanceRepositoryImpl.kt
```diff
@@ -16,24 +16,46 @@
...
             it[appId] = instance.appId
             it[buildId] = instance.buildId
             it[envId] = instance.envId
+            it[lastHeartbeatAt] = CurrentDateTime
+            it[status] = AgentHeartbeatStatus.RUNNING.name
+        }
+    }
+
+    override suspend fun updateHeartbeat(instance: InstanceHeartbeat) {
+        InstanceTable.update(where = {
```
RomanDavlyatshin, line 52: What if instance does not exist yet?

iryabov: When the agent starts and sends a PUT /instances request, the instance is assigned the RUNNING status and a last_heartbeat_at timestamp. If the scheduled heartbeat runs before the instance record has been created, the UPDATE simply affects zero rows and does not produce an error.

